Issue #23, opened 2017-12-17
All addresses should be with checksums


berndbohmeier:
The relay server returns now addresses with checksums (https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md#specification)
For consistency all addresses return by the clientlib should be with checksums now. So it needs to be implemented for the wallet, for example user.address

dakingha69:
Closed via #29

PR #29, opened 2018-01-09
Checksummed address

dakingha69:
Solves #23

### DIFF
--- a/src/CurrencyNetwork.ts
+++ b/src/CurrencyNetwork.ts
@@ -42,7 +42,7 @@ export class CurrencyNetwork {
   }
 
   public getDecimals (networkAddress: string, decimals?: number): Promise<any> {
-    if (!this.utils.checkAddress(networkAddress)) {
+    if (!ethUtils.isValidChecksumAddress(networkAddress)) {
       return Promise.reject(`${networkAddress} is not a valid address.`)
     }
     return Promise.resolve(
--- a/src/CurrencyNetwork.ts
+++ b/src/CurrencyNetwork.ts
@@ -11,21 +11,21 @@ export class CurrencyNetwork {
   }
 
   public getInfo (networkAddress: string): Promise<any> {
-    if (!ethUtils.isValidAddress(networkAddress)) {
+    if (!ethUtils.isValidChecksumAddress(networkAddress)) {
       return Promise.reject(`${networkAddress} is not a valid address.`)
     }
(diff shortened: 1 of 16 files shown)


src/CurrencyNetwork.ts
```diff
@@ -10,21 +11,21 @@ export class CurrencyNetwork {
   }
 
   public getInfo (networkAddress: string): Promise<any> {
-    if (!this.utils.checkAddress(networkAddress)) {
+    if (!ethUtils.isValidAddress(networkAddress)) {
```
berndbohmeier: Why do you not check for a checksum address here?

berndbohmeier:
And I think the clientlib should expose a function that checks if an address is valid. This can then be used by the mobileapp for validating input

src/CurrencyNetwork.ts
dakingha69: didnt want to make the check to strict originally.. but yea we should always check for checksum at this point

berndbohmeier: Yes maybe it is to strict. Can you make it similar to this https://github.com/ethereum/eth-utils/blob/master/eth_utils/address.py#L80
So only check for checksum address if there are mixed case letter? if all are lower case or all are upper case it was probably not intended to get checked for a checksum.

tests/e2e/Trustline.test.ts
```diff
@@ -117,8 +117,8 @@ describe('e2e', () => {
             setTimeout(() => {
               tl2.trustline.getUpdates(networkAddress).then(updates => {
                 const latestUpdate = updates[updates.length - 1]
-                expect(latestUpdate.address).to.equal(user1.address)
-                expect(latestUpdate.amount).to.equal(1500)
+                expect(latestUpdate.address.toLowerCase()).to.equal(user1.address.toLowerCase())
```
berndbohmeier: Did that slip in by accident? There should be no lowercase needed anymore right?

src/CurrencyNetwork.ts
```diff
@@ -10,21 +11,21 @@ export class CurrencyNetwork {
   }
 
   public getInfo (networkAddress: string): Promise<any> {
-    if (!this.utils.checkAddress(networkAddress)) {
+    if (!ethUtils.isValidChecksumAddress(networkAddress)) {
```
berndbohmeier: And please use the new created checkAddress function also here in the code, so it is the same everywhere

PR #39, opened 2018-01-30
Fix/eth fees

dakingha69:
- change format of `ethFees` and `maxFees` to `{ value, decimals, raw }`
- fix to use `estimatedGas` from relay server

### DIFF
--- a/src/Transaction.ts
+++ b/src/Transaction.ts
@@ -18,9 +18,10 @@ export class Transaction {
                      estimatedGas?: number): Promise<any> {
     return this.getTxInfos(userAddress)
       .then(txinfos => {
+        const gasLimit = estimatedGas ? estimatedGas * 1.5 : 2000000
         const txOptions = {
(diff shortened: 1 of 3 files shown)


src/Transaction.ts
```diff
@@ -14,7 +14,8 @@ export class Transaction {
                      contractAddress: string,
                      contractName: string,
                      functionName: string,
-                     parameters: any[]): Promise<any> {
+                     parameters: any[],
+                     estimatedGas?: number): Promise<any> {
     return this.getTxInfos(userAddress)
       .then(txinfos => {
         const txOptions = {
```
berndbohmeier: You need to return the ethFees according to the gasLimit you set. THis is the only reliable thing and determines the max eth spend. 
So ( in pseudocode):
```
gasLimit = estimatedGas ? estimatedGas * 1.2 : 2000000
txOptions.gasLimit = gasLimit
...
...
ethFees: this.utils.formatAmount(
            gasLimit * txOptions.gasPrice, 18
          )
```

PR #40, opened 2018-02-07
Improvement of buildUrl

### LINKED ISSUE #38: Escaping in buildUrl, createLink
We need to improve the encoding of parameters into the url and prevent problems when special characters like ? and & are included in the parameters. 

See
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/encodeURI
and
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

dakingha69:
Closes #38 
- merged buildUrl and createLink into one function
- parameters get encoded

### DIFF
--- a/src/Contact.ts
+++ b/src/Contact.ts
@@ -14,8 +14,8 @@ export class Contact {
 
   public createLink (address: string, username: string): Promise<string> {
     return new Promise((resolve, reject) => {
-      const params = [ address, username ]
-      resolve(this.utils.createLink('contact', params))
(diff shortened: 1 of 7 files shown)


dakingha69:
Ok reintroduced `createLink` for now. 

Regarding http I am not quite sure. I think @jensneuber should know more about it.

And to have something like `buildUrl(baseUrl: string, params: any[], optionalParams? any)` definitely makes sense. Will implement that at a later point.

PR #43, opened 2018-02-27
Unify prepare fx

### LINKED ISSUE #37: Unify prepare Functions
The prepare Functions in Payment and Transaction should be unified. I would propose this:

- In Payment use values (Euro, Ether) and in Transaction use raw values (cents, wei)
This would mean calculation from value to rawvalue will always happen in Payment and not in Transaction. So move EthToWei usage to Transaction and make sure that code that uses prepValueTransaction gets updated.

- Let prepareFuncTx also specify gasPrice and gasLimit like in prepareValueTx

dakingha69:
Closes #37

### DIFF
--- a/src/Payment.ts
+++ b/src/Payment.ts
@@ -24,34 +24,32 @@ export class Payment {
   }
 
   public prepare (
-    networkAddress: string,
-    receiver: string,
+    network: string,
+    to: string,
     value: number,
-    decimals: any = {},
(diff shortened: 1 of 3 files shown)


dakingha69:
Oh yes that should be only one which wasn't adjusted. The rest should be changed and tested. Instead of changing it to `prepareEth` I adjusted the `prepareValueTx` inputs accordingly.

PR #45, opened 2018-03-04
Add event stream and messaging over websockets

berndbohmeier:
The tests might be flaky. I dont like them. Maybe we can fix them that we dont require so many timeouts?

Steps for an event stream:
- Connect to Websocket, currently at `api/v1/streams/events`
- Sent subscription message
  - We use jsonrpc and the message to call is `subscribe` with keyword parameters `event`, the name of the event, currently only "all" as argument is supported, and `user`, which is the address of the user
- The result of this rpc call is a subscription id. The server will send a notification named "subscription_{id}" with the events for this subscription, so we need to register a function with this name and handle the incoming events. 
- return the incoming events to the stream 

How to use the stream:
- call subscribe(event => doSth(event)) and unsubscribe() on them


dakingha69:
looking good, works as expected. travis build is failing because in `relay/develop` the `feature/messaging` branch is not merged yet. waiting for PR https://github.com/trustlines-network/relay/pull/59 to merged.

PR #48, opened 2018-03-21
format events after fetch for all networks

jensneuber:


berndbohmeier:
What about also add an event for websocket close in the error handler?

PR #72, opened 2018-05-08
Default export main class

dakingha69:
Changes the export of the main class `TLNetwork` to default.


dakingha69:
Ah ok good points... My intention was originally to solve an other issue using `default` export but was going in the wrong direction. I will close this PR then and open a new one for the webpack update later.

PR #80, opened 2018-06-12
Fix/dependency vulnerabilites

dakingha69:
Fixes clientlib part of https://github.com/trustlines-network/project/issues/125


dakingha69:
true makes sense...

PR #92, opened 2018-06-14
Remove try catch and add async

### LINKED ISSUE #85: Remove try/catch blocks
Remove unnecessary try/catch blocks in async/await functions.

### LINKED ISSUE #86: Include async into functions with callbacks
Some functions still return `new Promise...` instead of using `async`. Refactor functions to use async.

dakingha69:
Closes #85 and closes #86 

- Removes unnecessary `try/catch` blocks
- Adds `async` to functions which were missed in the last refactoring
- Reintroduces `async createLink`

Note: In `User.ts` there are still a lot of functions which use `new Promise...`. They are necessary though because we need to wrap the callback based functions of `eth-lightwallet` to use them as promises.

### DIFF
--- a/src/CurrencyNetwork.ts
+++ b/src/CurrencyNetwork.ts
@@ -33,12 +33,8 @@ export class CurrencyNetwork {
    * @returns A network object with information about name, decimals, number of users and address.
    */
   public async getInfo (networkAddress: string): Promise<NetworkDetails> {
-    try {
-      await this._checkAddresses([networkAddress])
-      return this._utils.fetchUrl<NetworkDetails>(`networks/${networkAddress}`)
-    } catch (error) {
-      return Promise.reject(error)
-    }
+    await this._checkAddresses([networkAddress])
+    return this._utils.fetchUrl<NetworkDetails>(`networks/${networkAddress}`)
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/CurrencyNetwork.ts
```diff
@@ -114,27 +98,23 @@ export class CurrencyNetwork {
...
+    const networkAddresses = currencyNetworks.map(c => ethUtils.toChecksumAddress(c.address))
+    return networkAddresses.indexOf(ethUtils.toChecksumAddress(contractAddress)) !== -1
   }
 
   /**
    * Checks if given addresses are valid ethereum addresses.
    * @param addresses Array of addresses that should be checked.
    */
-  private _checkAddresses (addresses: string[]): Promise<boolean> {
+  private async _checkAddresses (addresses: string[]): Promise<boolean> {
```
jensneuber: didn't think of this before .. now that I see it, it makes perfect sense 😄

src/User.ts
```diff
@@ -313,7 +289,7 @@ export class User {
    * Contains username and address.
    * @param username Custom username.
    */
-  public createLink (username: string): string {
+  public async createLink (username: string): Promise<string> {
```
jensneuber: 👍

src/CurrencyNetwork.ts
```diff
@@ -114,27 +98,23 @@ export class CurrencyNetwork {
...
     for (let address of addresses) {
       if (!this._utils.checkAddress(address)) {
-        return Promise.reject(`${address} is not a valid address.`)
+        throw new Error(`${address} is not a valid address.`)
       }
     }
-    return Promise.resolve(true)
+    return
```
berndbohmeier: return true?
How does this pass the type checker?!

PR #93, opened 2018-06-14
Refactor getDecimals

### LINKED ISSUE #83: Refactor currencyNetwork.getDecimals
Current implementation of `getDecimals` is confusing. The function should always return the provided `decimals` if given without querying. Also it should not make two requests. Instead try to query decimals with one request and determine that a network does not exist through a request failure.

dohaki:
Closes #83 

- Changes `getDecimals` to be more clear
- Adds error test

### DIFF
--- a/src/CurrencyNetwork.ts
+++ b/src/CurrencyNetwork.ts
@@ -84,6 +84,9 @@ export class CurrencyNetwork {
       const network = await this._utils.fetchUrl<NetworkDetails>(`networks/${networkAddress}`)
       return network.decimals
     } catch (error) {
+      if (error.message.includes('Status 404')) {
+        throw new Error(`${networkAddress} seems not to be a network address. Decimals have to be explicit.`)
+      }
       throw error
     }
   }
--- a/tests/e2e/CurrencyNetwork.test.ts
+++ b/tests/e2e/CurrencyNetwork.test.ts
@@ -83,11 +83,9 @@ describe('e2e', () => {
 
       it('should throw error', async () => {
         const errMsg = [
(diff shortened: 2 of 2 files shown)


src/CurrencyNetwork.ts
```diff
@@ -76,21 +76,12 @@ export class CurrencyNetwork {
...
-      if ((typeof decimals === 'number') && decimals) {
-        return decimals
-      } else {
-        return Promise.reject(`${networkAddress} is a token address. Decimals have to be explicit.`)
-      }
+    if (decimals && typeof decimals === 'number') {
+      return decimals
     }
+    // TODO replace with list of known currency network in clientlib
+    const network = await this._utils.fetchUrl<NetworkDetails>(`networks/${networkAddress}`)
```
berndbohmeier: What will happen if the address is not a network? Before there was a meaningful error, can we have that back?

berndbohmeier: For example if a 404 comes back, we can throw an error like before.

dakingha69: The meaningful error is thrown here:
https://github.com/trustlines-network/clientlib/blob/3205002e761441689f843153709638e1f81a4204/src/Utils.ts#L52
So basically it returns the error message sent by the relay server. Shouldn't this suffice?

berndbohmeier: No, because the user of this function should not care about how you find out that it is not a network. Then the abstraction is broken. If we would check it differently maybe by comparing it to an internal list, the error should stay the same.

dakingha69: Ok got it. Will throw an error like before then

PR #94, opened 2018-06-14
Add return value to _checkAddresses

dakingha69:


dakingha69:
Moved this fix to PR #93

PR #96, opened 2018-06-19
Async/Await in Tests

### LINKED ISSUE #89: Use async/await in tests
To make the tests more readable we should also include async/await in our tests

dakingha69:
Closes #89 

- adds async/await in tests for better readability


tests/e2e/Event.test.ts
```diff
@@ -13,59 +13,70 @@ describe('e2e', () => {
...
-          tl2.trustline.prepareUpdate(networkAddress2, user1.address, 500, 1000)
+          tl1.trustline.prepareUpdate(network2.address, user2.address, 1000, 500),
+          tl2.trustline.prepareUpdate(network2.address, user1.address, 500, 1000)
         ])
         await Promise.all([
           tl1.trustline.confirm(txObj1.rawTx),
           tl2.trustline.confirm(txObj2.rawTx)
         ])
-        await new Promise(resolve => setTimeout(() => resolve(), 1000))
+        await wait()
```
jensneuber: is 2000 ms enough?

dakingha69: yes most of the time this should be enough :)

PR #105, opened 2018-07-06
Feature/exchange events

### LINKED ISSUE #58: Function to get event logs related to DEX contract
`exchange.getEvents(filter?)` function to retrieve all events related to a DEX contract. Possible events are `LogFill`, `LogError`, `LogCancel`. See https://github.com/0xProject/0x-monorepo/blob/6ba6bec1054ee89edc48d2128940c84c727a4b1d/packages/0x.js/src/contract_wrappers/exchange_wrapper.ts#L711

### LINKED ISSUE #116: E2E tests fail sometimes because of wrong decimals
The e2e tests sometimes fail with 
```
1) e2e
       Exchange
         #getOrderByHash()
           should return order by its hash:

      AssertionError: expected { Object (taker, availableTakerTokenAmount, ...) } to deeply equal { Object (exchangeContractAddress, expirationUnixTimestampSec, ...) }
      + expected - actual

         }
         "hash": "0xfadc271526c744c60e5f2e6586dfd8efad0f5def6170f7e704089e73cf353792"
         "maker": "0xf8E191d2cd72Ff35CB8F012685A29B31996614EA"
         "makerFee": {
      -    "decimals": 6
      +    "decimals": 2
           "raw": "0"
           "value": "0"
         }
         "makerTokenAddress": "0x3DD0864668C36D27B53a98137764c99F9FD5B7B2"
```
It seems that the selection of the currency networks is not reliable or so.

dakingha69:
Closes #58 and closes #116 

~NOTE: Waiting for https://github.com/trustlines-network/relay/pull/135 to be merged, so tests can succeed.~


tests/e2e/Exchange.test.ts
```diff
@@ -158,50 +161,139 @@ describe('e2e', () => {
...
-            done()
-          }))
-          .catch(e => done(e))
+      before(async () => {
+        order = await tl1.exchange.makeOrder(exchangeAddress, makerTokenAddress, takerTokenAddress, 1, 1)
+        const trustlines = await Promise.all([
+          tl2.trustline.getAll(makerTokenAddress),
+          tl2.trustline.getAll(takerTokenAddress)
+        ])
+        makerTLBefore = trustlines[0].find(tl => tl.address === tl1.user.address)
```
berndbohmeier: There is also an endpoint that you can query for exact that trustline.

tests/e2e/Event.test.ts
```diff
@@ -68,18 +74,99 @@ describe('e2e', () => {
...
+        const exTxIds = await Promise.all([
+          tl2.exchange.confirm(fillTx.rawTx),
+          tl1.exchange.confirm(cancelTx.rawTx)
         ])
+        fillTxId = exTxIds[0]
+        cancelTxId = exTxIds[1]
         await wait()
       })
 
+      it('should return all events', async () => {
```
berndbohmeier: Please also add a test that test the event more and that the keys are right, something like https://github.com/trustlines-network/clientlib/blob/3dc16342e6e9dfbbbd7b51c294438d71d8fa7eb4/tests/e2e/Event.test.ts#L252. 
Does not have to be here, but somewhere.

tests/e2e/Exchange.test.ts
```diff
@@ -158,50 +161,139 @@ describe('e2e', () => {
...
+      })
+
+      it('should return LogFill events', async () => {
+        const events = await tl1.exchange.getLogs(exchangeAddress, { type: 'LogFill' })
+        const fillEvent = events.find(e => e.orderHash === order.hash)
+        expect(fillEvent).to.have.keys(fillEventKeys)
+        expect(fillEvent.transactionId).to.equal(fillTxId)
+      })
+
+      it('should return LogCancel events', async () => {
```
berndbohmeier: Please check that you only get logCancel events here too.

tests/e2e/Exchange.test.ts
```diff
@@ -158,50 +161,139 @@ describe('e2e', () => {
...
+        const fillEvent = filteredEvents.find(e => e.type === 'LogFill')
+        const cancelEvent = filteredEvents.find(e => e.type === 'LogCancel')
+        expect(filteredEvents).to.have.length(2)
+        expect(fillEvent).to.have.keys(fillEventKeys)
+        expect(fillEvent.transactionId).to.equal(fillTxId)
+        expect(cancelEvent).to.have.keys(cancelEventKeys)
+        expect(cancelEvent.transactionId).to.equal(cancelTxId)
+      })
+
+      it('should return LogFill events', async () => {
```
berndbohmeier: Please check that you only get logFill events here too. Either by the number, or by the type or something like that.

src/Exchange.ts
```diff
@@ -332,6 +276,21 @@ export class Exchange {
...
+    filter: EventFilterOptions = {}
+  ): Promise<AnyExchangeEvent[]> {
+    const { _event, _user, _utils } = this
+    const baseUrl = `exchange/${exchangeAddress}/users/${_user.address}/events`
+    const parameterUrl = _utils.buildUrl(baseUrl, filter)
+    const rawEvents = await _utils.fetchUrl<AnyExchangeEventRaw[]>(parameterUrl)
+    const formattedEvents = await _event.setDecimalsAndFormat(rawEvents)
+    return formattedEvents.map(e => ({
+      ...e,
+      orderHash: e.orderHash.toLowerCase()
```
berndbohmeier: is that a bug in the relay we should fix? Should it be lower case? Then please open an issue.

tests/e2e/Event.test.ts
dakingha69: alright done

tests/e2e/Exchange.test.ts
dakingha69: done

tests/e2e/Exchange.test.ts
dakingha69: done

src/Exchange.ts
dakingha69: ah no it was a bug but is fixed now with the schemas in the relay. removed this

PR #108, opened 2018-07-10
Decimal strings to hex strings

### LINKED ISSUE #91: Big numbers for values
The current version of eth-lightwallet does not allow to pass BigNumbers or strings as values for value transactions. In the clientlib we do something like
```javascript
const txOptions = {
  ...
  value: new BigNumber(rawValue).toNumber()
}
```
which omits the digits that occur after 15 digits. We have to find a way to not omit these digits.

### LINKED ISSUE #107: Parse decimal strings to hex strings before giving to eth-lightwallet
As eth-lightwallet only can handle number values or their hex string representations, we should always parse them to hex strings. I.e. for `value`, `gasPrice` and `gasLimit` when creating transactions. This would also cover #91.

dakingha69:
Closes #107 and closes #91.

### DIFF
--- a/src/Exchange.ts
+++ b/src/Exchange.ts
@@ -268,7 +268,11 @@ export class Exchange {
         ecSignature.s
       ], {
         gasPrice,
-        gasLimit: gasLimit || (takerPathObj.estimatedGas + makerPathObj.estimatedGas) * 1.5
+        gasLimit: gasLimit ||
+          new BigNumber(takerPathObj.estimatedGas)
+            .plus(new BigNumber(makerPathObj.estimatedGas))
+            .multipliedBy(1.5)
+            .toString()
       }
     )
     return {
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -114,22 +114,22 @@ export class Utils {
 
   /**
    * Returns the smallest representation of a number.
-   * @param value representation of number in biggest unit
-   * @param decimals number of decimals
+   * @param value Representation of number in biggest unit.
+   * @param decimals Number of decimals.
    */
   public calcRaw (value: number | string, decimals: number): string {
-    const x = new BigNumber(value)
-    return x.times(Math.pow(10, decimals)).toString()
+    const factor = new BigNumber(10).exponentiatedBy(decimals)
+    return new BigNumber(value).multipliedBy(factor).toString()
   }
 
   /**
    * Returns the biggest representation of a number.
-   * @param raw representation of number in smallest unit
-   * @param decimals nuber of decimals
+   * @param raw Representation of number in smallest unit.
+   * @param decimals Number of decimals.
    */
   public calcValue (raw: number | string, decimals: number): string {
-    const x = new BigNumber(raw)
-    return x.div(Math.pow(10, decimals)).toString()
+    const divisor = new BigNumber(10).exponentiatedBy(decimals)
+    return new BigNumber(raw).dividedBy(divisor).toString()
   }
 
   /**
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -205,4 +205,12 @@ export class Utils {
     const wei = new BigNumber(1000000000000000000)
     return eth.times(wei).toNumber()
   }
+
+  /**
+   * Returns the hexdecimal representation of given decimal string.
+   * @param decimalStr Decimal string representation of number.
+   */
+  public convertDecToHex (decimalStr: string): string {
+    return new BigNumber(decimalStr).toString(16)
+  }
 }
--- a/src/typings.ts
+++ b/src/typings.ts
@@ -27,9 +27,9 @@ export interface TLNetworkConfig {
 }
 
 export interface TxOptions {
-  value?: number | string,
-  gasPrice?: number,
-  gasLimit?: number
+  value?: string,
+  gasPrice?: string,
+  gasLimit?: string
 }
 
 export interface TLOptions extends TxOptions {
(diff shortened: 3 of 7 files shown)


src/Utils.ts
```diff
@@ -205,4 +205,12 @@ export class Utils {
     const wei = new BigNumber(1000000000000000000)
     return eth.times(wei).toNumber()
   }
+
+  /**
+   * Returns the hexdecimal representation of given decimal string.
+   * @param decimalStr Decimal string representation of number.
+   */
+  public convertDecToHex (decimalStr: string): string {
```
berndbohmeier: can you add some tests please. Does it also work for numbers or hex strings? If not what will happen?

src/Exchange.ts
```diff
@@ -268,7 +268,11 @@ export class Exchange {
         ecSignature.s
       ], {
         gasPrice,
-        gasLimit: gasLimit || (takerPathObj.estimatedGas + makerPathObj.estimatedGas) * 1.5
+        gasLimit: gasLimit ||
+          new BigNumber(takerPathObj.estimatedGas)
+            .plus(new BigNumber(makerPathObj.estimatedGas))
+            .multipliedBy(1.5)
+            .toString()
```
berndbohmeier: I think we should use BigNumbers internally and only convert it to a string when exporting it out of the lib. That means we can always do calculation. Otherwise we always need to convert.

src/typings.ts
```diff
@@ -127,11 +127,11 @@ export interface TxInfos {
   /**
    * Amount of ETH in gwei for every unit of gas user is willing to pay
    */
-  gasPrice: number,
+  gasPrice: string,
```
berndbohmeier: As I said I think this should be BigNumber

src/typings.ts
```diff
@@ -27,9 +27,9 @@ export interface TLNetworkConfig {
 }
 
 export interface TxOptions {
-  value?: number | string,
-  gasPrice?: number,
-  gasLimit?: number
+  value?: string,
+  gasPrice?: string,
```
berndbohmeier: Think it should be BigNumber

src/Utils.ts
dakingha69: Ok done. Works also for numbers. Hex strings only work if prefixed with 0x.

src/Utils.ts
```diff
@@ -114,22 +114,22 @@ export class Utils {
...
-   * @param decimals nuber of decimals
+   * @param raw Representation of number in smallest unit.
+   * @param decimals Number of decimals.
    */
-  public calcValue (raw: number | string, decimals: number): string {
-    const x = new BigNumber(raw)
-    return x.div(Math.pow(10, decimals)).toString()
+  public calcValue (raw: number | string | BigNumber, decimals: number): string {
+    const divisor = new BigNumber(10).exponentiatedBy(decimals)
+    return new BigNumber(raw).dividedBy(divisor).toString()
```
berndbohmeier: Internal functions should return BigNumbers from now

src/Utils.ts
```diff
@@ -114,22 +114,22 @@ export class Utils {
...
    * Returns the smallest representation of a number.
-   * @param value representation of number in biggest unit
-   * @param decimals number of decimals
+   * @param value Representation of number in biggest unit.
+   * @param decimals Number of decimals.
    */
-  public calcRaw (value: number | string, decimals: number): string {
-    const x = new BigNumber(value)
-    return x.times(Math.pow(10, decimals)).toString()
+  public calcRaw (value: number | string | BigNumber, decimals: number): string {
```
berndbohmeier: Internal functions should return BigNumbers from now.

src/Payment.ts
```diff
@@ -98,11 +100,15 @@ export class Payment {
     value: number | string,
     options: PaymentOptions = {}
   ): Promise<TxObject> {
+    const { gasLimit, gasPrice } = options
     return this._transaction.prepValueTx(
       this._user.address,
       receiverAddress,
       this._utils.calcRaw(value, 18),
```
berndbohmeier: As this will be a BigNumber after the change, we have to call convertToHex here then

berndbohmeier: Same in Trustline.js

src/typings.ts
```diff
@@ -26,10 +26,19 @@ export interface TLNetworkConfig {
   wsProtocol?: string
 }
 
+/**
+ * For internal use of `prepFuncTx` and `prepValueTx`.
+ */
+export interface RawTxOptions {
```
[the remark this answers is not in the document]
dakingha69: done

src/Utils.ts
```diff
@@ -205,4 +205,12 @@ export class Utils {
     const wei = new BigNumber(1000000000000000000)
     return eth.times(wei).toNumber()
   }
+
+  /**
+   * Returns the hexdecimal representation of given decimal string.
+   * @param decimalStr Decimal string representation of number.
+   */
+  public convertDecToHex (decimalStr: string | number | BigNumber): string {
+    return new BigNumber(decimalStr).toString(16)
```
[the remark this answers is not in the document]
dakingha69: done

src/Utils.ts
```diff
@@ -205,4 +205,12 @@ export class Utils {
     const wei = new BigNumber(1000000000000000000)
     return eth.times(wei).toNumber()
   }
+
+  /**
+   * Returns the hexdecimal representation of given decimal string.
+   * @param decimalStr Decimal string representation of number.
+   */
+  public convertDecToHex (decimalStr: string | number | BigNumber): string {
```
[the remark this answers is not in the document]
dakingha69: done

PR #109, opened 2018-07-11
Package Updates

dakingha69:
- package updates
- lint fixes


berndbohmeier:
Will it still minimize the size of the bundle?

dakingha69:
Yes it is for `production` mode in the new webpack version. Also way smaller now: ~ 1.4 MB (still big though)

PR #114, opened 2018-07-17
Fix setup file for e2e tests to work with new deploy commandline

berndbohmeier:
Fixes the tests to work with new deploy script, see https://github.com/trustlines-network/contracts/pull/111

### DIFF
--- a/tests/e2e/Exchange.test.ts
+++ b/tests/e2e/Exchange.test.ts
@@ -30,9 +30,9 @@ describe('e2e', () => {
         tl1.exchange.getExAddresses(),
         tl1.currencyNetwork.getAll()
       ])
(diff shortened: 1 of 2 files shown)


dakingha69:
The problem is that `currencyNetwork.getAll` returns Currency Network only with high-level infomation:
```
{
  abbreviation: ...,
  name: ...,
  address: ....
}
```
Either we include decimals into the endpoint `/networks` or we randomly pick networks for the tests.

berndbohmeier:
Is it important that they have 2 decimals?

dakingha69:
No not really... We can then just take the first 2 networks in the test for now.

berndbohmeier:
Perfect. Lgtm

PR #115, opened 2018-07-23
Include users own address in events

berndbohmeier:
Fix tests as the users own address is now also included in the events and the name from the counter party address was changed from address to counterParty
Needed because of https://github.com/trustlines-network/relay/pull/149


dakingha69:
Travis build should succeed when PR https://github.com/trustlines-network/clientlib/pull/105 is merged because #116 should also be fixed with https://github.com/trustlines-network/clientlib/pull/105.

Issue #119, opened 2018-07-26
Improve Tests


dakingha69:
### Problem
The current tests are a bit hard to read, too dependent on each other and incomplete. 

### Solution
- [ ] Improve readability by replacing `expect` with `assert`
- [ ] Increase independency between tests by using mocked classes/functions
- [ ] Add proper unit tests for every class
- [ ] Add proper integration tests for dependent classes
- [ ] Add negative tests
- tbd

PR #120, opened 2018-07-26
Web3 integration for signing tx

### LINKED ISSUE #100: Implement configurable web3 provider

### LINKED ISSUE #102: Enhance / add feature to sign tx method
This should be done to use the configured web3 provider

documention should be updated

dakingha69:
Closes #100 and closes #102 

Known issues:
- Only rudimentary tests implemented, remaining tests will be implemented via issue #119 
- Wallet related methods not implemented in Web3Signer -> #122


src/signers/LightwalletSigner.ts
```diff
@@ -0,0 +1,87 @@
...
+// declare let lightwallet
+
+/**
+ * The LightwalletSigner class contains functions for signing transactions with eth-lightwallet.
+ */
+export class LightwalletSigner implements TxSigner {
+  private _user: User
+  private _utils: Utils
+
+  constructor (user: User, utils: Utils) {
```
berndbohmeier: This is the wrong dependency direction. User should depend on the signer, not the other way arround. If a user wants to sign, he needs a signing service, not the other way around. Then you should not have the problem why you needed to refactor the onboarding stuff.

src/User.ts
```diff
@@ -84,11 +84,19 @@ export class User {
     }
   }
 
+  public async loadWeb3Account (): Promise<void> {
+    if (this._web3.eth.defaultAccount) {
+      this.address = this._web3.eth.defaultAccount
```
berndbohmeier: the address belongs to the signing service. So address should probably just forward to txSigner.address.

berndbohmeier: Then user does not have a direct dependency of web3.

src/signers/LightwalletSigner.ts
```diff
@@ -0,0 +1,87 @@
...
+      rlpTx = lightwallet.txutils.functionTx(
+        rawTx.functionCallData.abi,
+        rawTx.functionCallData.functionName,
+        rawTx.functionCallData.args,
+        txOptions
+      )
+    } else {
+      rlpTx = lightwallet.txutils.valueTx(txOptions)
+    }
+    const signedTx = await this._user.signTx(rlpTx)
```
berndbohmeier: Asking back to user here does not make sense. This is the signing service, so it should do the signing.

src/signers/TxSigner.ts
```diff
@@ -0,0 +1,12 @@
+import {
+  TxInfos,
+  RawTxObject
+} from '../typings'
+
+/**
+ * Interface for different signer strategies.
+ */
+export interface TxSigner {
```
berndbohmeier: Add address here

dakingha69: done

src/User.ts
```diff
@@ -67,124 +60,26 @@ export class User {
...
-          this.keystore,
-          pwDerivedKey,
-          rawTx,
-          this.address.toLowerCase() // NOTE eth-lightwallet does not handle checksum addresses
-        ))
-      })
-    })
+    const loadedAccount = await this._signer.loadAccount(serializedKeystore)
+    this.address = loadedAccount.address
+    this.keystore = loadedAccount.keystore
```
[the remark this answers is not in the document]
dakingha69: Yes true this is eth-lightwallet only... Will remove it. But it should still be returned by `loadAccount` and `createAccount` right? Because the client needs to store it.

src/User.ts
```diff
@@ -220,64 +97,76 @@ export class User {
...
+   * Returns a shareable link, which can be opened by other users who already have ETH
+   * and are willing to send some of it to the new user. The function is called by a
+   * new user who wants to get onboarded, respectively has no ETH or trustline.
+   * @param username Name of new user who wants to get onboarded.
+   * @param serializedKeystore Serialized [eth-lightwallet](https://github.com/ConsenSys/eth-lightwallet)
+   *                           keystore of new user who wants to get onboarded.
+   */
+  public async createOnboardingMsg (
+    username: string,
+    serializedKeystore: string
```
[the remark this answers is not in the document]
dakingha69: Still a relict from a previous version. Should use the loaded one... Will open an issue

src/User.ts
```diff
@@ -26,39 +24,34 @@ export class User {
...
-    this.address = address
-    this.keystore = keystore
-    this.pubKey = pubKey
-    return {
-      address: this.address,
-      keystore: this.keystore.serialize(),
-      pubKey: this.pubKey
-    }
+    const createdAccount = await this._signer.createAccount()
+    this.address = createdAccount.address
```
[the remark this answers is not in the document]
dakingha69: yes can do that

src/User.ts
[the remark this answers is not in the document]
berndbohmeier: okay.

src/User.ts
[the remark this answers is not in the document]
berndbohmeier: Yes, as it would otherwise break old code. But maybe we should open an issue to think about changing it.

PR #127, opened 2018-08-03
Fix build stages

berndbohmeier:


berndbohmeier:
Close and reopen because travis seems to hang.

PR #130, opened 2018-08-07
Remove npm publish on develop without tags

dakingha69:
I had to change the release/versioning flow a bit because there are some problems with the npm dist-tag `next` which we wanted to use for the latest commit on the `develop` branch. npm does not allow to update a already published version. So in order to have the latest develop branch with the tag `next` we would have to increment the version in the package.json somehow.

For a quick fix I would just manually bump the version via `npm version major | minor | patch` for official releases and only have the dist-tag `latest` as introduced in this PR.

Need some input on this @jensneuber and @berndbohmeier. Especially from you Jens as you were the one who wanted the `next` tag.


berndbohmeier:
So the problem is not the next tag, but that we do not change the version number right?

berndbohmeier:
We could also try something like https://www.npmjs.com/package/fluid-publish

berndbohmeier:
This would be more close how we do it with the python code.

PR #132, opened 2018-08-16
Fluid Publish on NPM

### LINKED ISSUE #131: Fluid publish on npm
As mentioned in PR https://github.com/trustlines-network/clientlib/pull/130, we only have the dist-tag `latest`  with git tags `major.minor.patch` for official releases where the publish on npm is triggered manually.

To be consistent to the versioning scheme/flow of the relay server, we should also automatically publish a `dev` version which should always be the most recent commit on the develop branch. This could be done with something like  https://www.npmjs.com/package/fluid-publish and a custom deploy script for travis.

dakingha69:
Closes #131 

- publishes most recent develop commit with tag ` X.x.x-dev.yyyymmddThhmmssZ.shortHash`
- can be downloaded via `npm i trustlines-clientlib@dev`


.travis.yml
```diff
@@ -44,3 +52,5 @@ notifications:
       - secure: "R7leHfENzyd6mAIdDS1JUmwZnvuG8QSgQfcjH6QGzh7c1UlUJHrP+u5G5fiWGB4mBNQexh1nOpW8+JlYFyEnWrXUgg2wXDhLxM45jaVmgpYmusNvLC+fk70o4k6KksSYOkoo2pxcthlsp2BqCfkVA8f1dxy7Q6RxQBp4KBoWfh+qQ00fCp/OsL/UuPlYz5qWohxAVyulIVVfvEWlplH870nE/VZD+K91KIB6CDBR1REjx4ZBq904jsD4sg4cYua7Rik6F1grfA84Lt68s2gG89d11qhMtkEE6ROU++N3ziOFNbQIWVMTHW0JqcCLZ+cbvLH48m3IeOhttkdjE0JbzK3WsmYlHlxkZGmqyxucYxAd9nODkxKRjcrc+OFKgQ359HyWBwf9AtHi6mPhrMGrOEcTPxNi+ZGpnYQgqlJ+akW2cXO/fcw6bLYU+i535MgGG6oqotkfS6k7IkZBnkw3ly9AwJ1n9XUHALKF21GtcCMGdSkUtQtnLUFxO0Wg7GIbMXxDXrLw8CF56u6AQy75F+/5LDTdfLHWgZKO+G5HrHUsFuq6S6EydEH2WxVYaJ7XS5M8TIDzEr8ZQdtTG0LJA+Yb3YxzcPbIRomQrWIw8MXdk3eA5MjIZy0cwXMdNeqSgFpTba1Yg98HQfjjW78QV5VBAxXK+iv1T7oooChu51s="
     on_success: change
     on_failure: always
+env:
```
berndbohmeier: if you only need this env variable in the one stage, please attach it directly to the stage. See https://github.com/trustlines-network/relay/blob/develop/.travis.yml#L48

dakingha69: Actually needs to be there because I required `always-auth = true` in the `.npmrc` file. So in the test stage we also need it to be able to do `npm install`.

PR #133, opened 2018-08-18
max transferable amount

jensneuber:
adds methods to get max transferable amount of 
- adjacent 
and 
- non-adjacent users

https://github.com/trustlines-network/project/issues/161

### DIFF
--- a/src/Payment.ts
+++ b/src/Payment.ts
@@ -156,10 +156,10 @@ export class Payment {
       value: this._utils.calcRaw(value, decimals).toString()
     }
     if (maximumFees) {
-      data['maxFees'] = maximumFees
+      data[ 'maxFees' ] = maximumFees
     }
     if (maximumHops) {
-      data['maxHops'] = maximumHops
+      data[ 'maxHops' ] = maximumHops
     }
     const endpoint = `networks/${networkAddress}/path-info`
     const { estimatedGas, fees, path } = await _utils.fetchUrl<PathRaw>(endpoint, {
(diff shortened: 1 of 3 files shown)


src/Payment.ts
```diff
@@ -212,4 +212,81 @@ export class Payment {
...
+   * @param networkAddress
+   *
+   * @return {Promise<{amount: Amount}>}
+   */
+  public async getMaxAmount (
+    networkAddress: string
+  ): Promise<any> {
+    const decimals = await this._currencyNetwork.getDecimals(networkAddress)
+    const userAddress = this._user.address
+    const endpoint = `networks/${networkAddress}/users/${userAddress}/spendable`
```
berndbohmeier: This endpoint just gives the available and is deprecated because of the confusing name, as it does not necessarily is the spendable amount because of the fees.

src/Payment.ts
```diff
@@ -212,4 +212,81 @@ export class Payment {
...
+   *
+   * @return {Promise<{amount: Amount}>}
+   */
+  public async getMaxAmountToUser (
+    networkAddress: string,
+    receiverAddress: string
+  ): Promise<any> {
+    const decimals = await this._currencyNetwork.getDecimals(networkAddress)
+    const userAddress = this._user.address
+    const endpoint = `networks/${networkAddress}/users/${userAddress}/spendables/${receiverAddress}`
```
berndbohmeier: This is also deprecated as it also does not take into account the fees. max-capacity-path-info can do this.

src/Payment.ts
berndbohmeier: I actually thought I removed it already.

berndbohmeier: I opened an issue. The docs are up to date with the source code.. But not with the project.

PR #137, opened 2018-09-25
Fix/end2end

schmir:
This makes running the end2end tests possible again. See https://github.com/trustlines-network/relay/pull/204


berndbohmeier:
I had to switch to a stable parity to fix the e2e test. This means this might get a problem again, if they dont fix the problem until they release the next version.

PR #138, opened 2018-09-26
LightwalletSigner Unit Tests

dakingha69:
Adds unit tests for LightwalletSigner.ts. Also some smaller refactoring which came up when writing the tests.

### DIFF
--- a/src/signers/LightwalletSigner.ts
+++ b/src/signers/LightwalletSigner.ts
@@ -361,7 +361,7 @@ export class LightwalletSigner implements TxSigner {
 
   private _assertUserLoaded (): boolean {
     if (!this.address || !this.pubKey || !this.keystore) {
-      throw new Error('No user loaded.')
+      throw new Error('No account/keystore loaded.')
     }
     return
   }
--- a/src/signers/LightwalletSigner.ts
+++ b/src/signers/LightwalletSigner.ts
@@ -295,14 +295,14 @@ export class LightwalletSigner implements TxSigner {
    * @param seed (optional) 12 word seed string
    */
   private _generateKeys (seed?: string): Promise<UserObject> {
-    return new Promise(resolve => {
+    return new Promise((resolve, reject) => {
       this._lightwallet.keystore.createVault({
         password: this._password,
         seedPhrase: seed || this._lightwallet.keystore.generateRandomSeed(),
         hdPathString: this._signingPath
       }, (err: any, keystore: any) => {
         if (err) {
-          throw err
+          return reject(err)
         }
         const userObj = this._getUserObject(keystore)
         resolve(userObj)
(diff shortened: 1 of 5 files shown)


src/signers/LightwalletSigner.ts
```diff
@@ -144,18 +144,20 @@ export class LightwalletSigner implements TxSigner {
    * @param theirPubKey Public key of sender of message.
    */
   public decrypt (encMsg: any, theirPubKey: string): Promise<any> {
-    return new Promise((resolve, reject) => {
+    this._assertUserLoaded()
+    return new Promise(resolve => {
       this.keystore.keyFromPassword(this._password, (err: any, pwDerivedKey: any) => {
         if (err) {
-          return reject(err)
+          throw err
```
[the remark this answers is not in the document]
dakingha69: ahh ok... now I understand. I will change this.

tests/helpers/FakeEthLightwallet.ts
```diff
@@ -0,0 +1,294 @@
...
+  }
+  private _createVault (options, callback) {
+    if (this.errors['createVault']) {
+      return callback(new Error('Mocked error in lightwallet.keystore.createVault()'), undefined)
+    }
+    return callback(undefined, this.keystore)
+  }
+  private _generateRandomSeed () {
+    if (this.errors['generateRandomSeed']) {
+      throw new Error('Mocked error in lightwallet.keystore.generateRandomSeed()')
```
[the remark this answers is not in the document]
dakingha69: I wanted to make sure that every possible error thrown by eth-lightwallet is properly caught. I looked into the source code of `generateRandomSeed()` and it can only throw an error when it is called with an optional argument `extraEntropy`. But as we don't use the function with an argument, it doesn't really make sense to expect an error here... Will remove it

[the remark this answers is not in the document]
berndbohmeier: Okay, then you would need to try and catch it and then call reject with the error.

src/signers/LightwalletSigner.ts
```diff
@@ -105,6 +114,9 @@ export class LightwalletSigner implements TxSigner {
    * Returns ETH balance of loaded user by querying the relay server.
    */
   public async getBalance (): Promise<Amount> {
+    if (!this._isUserLoaded()) {
+      return Promise.reject(this._userNotLoadedError)
```
berndbohmeier: here I would use throw, as this is one of the reasons why you want to use async/await so that you can write it as usual.

tests/helpers/FakeEthLightwallet.ts
```diff
@@ -0,0 +1,293 @@
...
+    setTimeout(() => callback(undefined, this._pwDerivedKey))
+    return
+  }
+  private _exportPrivateKey (address, pwDerivedKey) {
+    if (this.errors['exportPrivateKey']) {
+      throw new Error('Mocked error in lightwallet.keystore.exportPrivateKey')
+    }
+    return this._privateKey
+  }
+  private _createVault (options, callback) {
```
berndbohmeier: Can you do the same thing with the setTimeout here to be sure we also are save here?

src/signers/LightwalletSigner.ts
berndbohmeier: Rule of thumb, if you dont have reject, but need too use Promise.reject, you can probably do it just with throw.

dakingha69: ah yes makes sense.. good rule of thumb will remember it

tests/helpers/FakeEthLightwallet.ts
dakingha69: yes done

berndbohmeier:
The error is caused by a parity update, so unrelated to this.

PR #144, opened 2018-11-13
Support interest rates

### LINKED ISSUE #141: Support interest rates
Interest rates are soon to be implemented in the contracts and the relay server:
- https://github.com/trustlines-network/contracts/pull/128
- https://github.com/trustlines-network/relay/pull/201

The clientlib should also be modified to support interest rates.

Related user story https://github.com/trustlines-network/project/issues/253

dakingha69:
Closes #141 

NOTE: Waiting for related relay PR https://github.com/trustlines-network/relay/pull/201 to be merged so tests can succeed.


src/CurrencyNetwork.ts
```diff
@@ -56,33 +66,45 @@ export class CurrencyNetwork {
...
   }
 
   /**
-   * Returns the decimals specified in a currency network.
+   * Returns the network decimals and interest decimals specified in a currency network.
    * @param networkAddress Address of currency network.
    * @param decimals If decimals are known they can be provided manually.
    */
-  public async getDecimals (networkAddress: string, decimals?: number): Promise<number> {
+  public async getDecimals (
```
berndbohmeier: I think that this function is weird. That it will if you give it some decimals will just return them. This also confused Thore during the hackathon. I think this function should just return the decimals or none if it cant find one, and the caller will need to decide whether to use this function or nor. 
I will open a refactor issue if you agree.

src/CurrencyNetwork.ts
```diff
@@ -56,33 +66,45 @@ export class CurrencyNetwork {
...
+    networkAddress: string,
+    decimalsOptions: DecimalsOptions = {}
+    ): Promise<DecimalsObject> {
+    const { networkDecimals, interestRateDecimals } = decimalsOptions
+    const decimalsObject = { networkDecimals, interestRateDecimals }
```
berndbohmeier: Also what was your reason to combine those two and not have two functions getDecimals, getInterestDecimals?

src/Trustline.ts
```diff
@@ -41,16 +41,20 @@ export class Trustline {
...
-   *                 i.e. 1.23 if network has to 2 decimals.
-   * @param options Options for creating an ethereum transaction. See type `TLOptions` for more information.
-   * @param options.decimals Decimals of currency network can be provided manually if known.
+   * @param creditlineGiven Proposed creditline limit given by initiator to counterparty,
+   *                        i.e. 1.23 if network has to 2 decimals.
+   * @param creditlineReceived Proposed creditline limit received by initiator from counterparty,
+   *                           i.e. 1.23 if network has to 2 decimals.
+   * @param options Options for creating an `updateTrustline` ethereum transaction.
+   *                See type `TrustlineUpdateOptions` for more information.
+   * @param options.interestRateGiven Proposed interest rate given by initiator to counterparty in % per year.
```
berndbohmeier: What is the reason to add this to the option object instead of having the interests as an own argument along with creditlineGiven? I think it would be cleaner as a normal argument with a default value.

src/CurrencyNetwork.ts
dakingha69: Yes please do so. The function always felt a bit hacky also in my eyes...

src/CurrencyNetwork.ts
dakingha69: I also thought about this alternative. But I wanted to keep the number of requests as small as possible because we retrieve the decimals currently via the relay server. If we have two functions we would need to call the same request twice...

But I also see the point that splitting it up is cleaner. Also `getInterestRateDecimals` would only be needed for `CurrencyNetwork` and `Trustline` related functions :thinking:

src/Trustline.ts
dakingha69: Here I also thought about this. So you mean something like
```javascript
public async prepareUpdate(
  networkAddress,
  counterpartyAddress,
  clGiven,
  clReceived,
  interestGiven = 0,
  interestReceived = 0,
  options = {}
) {...}
```
The problem would be then that someone could not call something like
```javascript
prepareUpdate(
  '0x...',
  '0x..',
  1000,
  2000,
  {
    // options
    ...
  }
)
```

src/CurrencyNetwork.ts
berndbohmeier: Opened and issue

src/CurrencyNetwork.ts
berndbohmeier: Makes sense, lets keep it like this for now.

src/Trustline.ts
berndbohmeier: I think someone should anyway never do the second options, as it is confusing what is going on and the proper way with naming the argument should still work:
```
prepareUpdate(
  '0x...',
  '0x..',
  1000,
  2000,
  options: {
    // options
    ...
  }
)
```
Or is that not possible in javascript?

dakingha69: I am afraid that named arguments like this is not possible in Javascript...

berndbohmeier: Ah okay, then I guess this was the right decision if we want to have interests optional.

PR #145, opened 2018-11-14
Close trustline

### LINKED ISSUE #142: Close trustlines
## Description
Functionality to close a trustline should be implemented as supported by the relay server.

## Tasks
- [ ] Add functionality to clientlib
- [ ] Add tests

## Related stories / issues
- https://github.com/trustlines-network/project/issues/244

dakingha69:
Closes #142


src/Trustline.ts
```diff
@@ -216,6 +220,127 @@ export class Trustline {
...
+            ? new BigNumber(gasLimit)
+            : new BigNumber(estimatedGas).multipliedBy(1.5).integerValue()
+        }
+      )
+
+      return {
+        rawTx,
+        path,
+        maxFees,
+        value,
```
berndbohmeier: Im not really sure what this value is here. I think we can not really know what value will be transferred. This is exactly the point of the closeTrustline method, that it will transfer a variable amount to be able to close it. If this is the value from closePath, we should not return it here, because this might not be the actual value that is transferred.

src/Trustline.ts
```diff
@@ -216,6 +220,127 @@ export class Trustline {
...
 
+  /**
+   * Prepares an ethereum transaction object for closing a trustline.
+   * @param networkAddress Address of a currency network.
+   * @param counterpartyAddress Address of counterparty to who the trustline should be settled.
+   * @param options Payment options. See `PaymentOptions` for more information.
+   * @param options.decimals Decimals of currency network can be provided manually.
+   * @param options.maximumHops Max. number of hops for transfer.
+   * @param options.maximumFees Max. transfer fees user if willing to pay.
+   */
```
berndbohmeier: Can we describe the return value too?

src/Trustline.ts
```diff
@@ -216,6 +220,127 @@ export class Trustline {
...
+  /**
+   * Returns a path for closing a trustline between sender and counterparty.
+   * @param networkAddress Address of a currency network.
+   * @param senderAddress Address of sender.
+   * @param counterpartyAddress Address of counterparty of trustline.
+   * @param options Payment options. See `PaymentOptions` for more information.
+   * @param options.networkDecimals Decimals of currency network can be provided manually.
+   * @param options.maximumHops Max. number of hops for transfer.
+   * @param options.maximumFees Max. transfer fees user if willing to pay.
+   */
```
berndbohmeier: Document return value

src/Trustline.ts
```diff
@@ -216,6 +220,127 @@ export class Trustline {
...
+        headers: new Headers({ 'Content-Type': 'application/json' }),
+        body: JSON.stringify(data)
+      }
+    )
+
+    return {
+      path,
+      estimatedGas: new BigNumber(estimatedGas),
+      maxFees: _utils.formatToAmount(fees, decimals.networkDecimals),
+      value: _utils.formatToAmount(value, networkDecimals)
```
berndbohmeier: Here is is fine to return the value, as it is the current estimation of what should be transferred if using a normal transfer. But this should be documented.

tests/e2e/Trustline.test.ts
```diff
@@ -385,5 +392,108 @@ describe('e2e', () => {
...
+        expect(closeTx.value).to.have.keys(['raw', 'value', 'decimals'])
+        expect(closeTx.value.value).to.equal(transferAmount.toString())
+      })
+
+      it('should not be possible to prepare a close in the wrong direction.', async () => {
+        // Send the prepare settle to the relay, expecting that a 501 HTTP status code get responded.
+        try {
+          await tl2.trustline.prepareClose(networkWithoutInterestRates.address, tl1.user.address)
+        } catch (err) {
+          expect(err).to.be.an('error')
```
berndbohmeier: I dont think we should have this test here. This is currently not implemented, but should be in the future. This reads as if we would expect it from beeing not working. 
I would say we either remove the test or make it a proper test but skip it for now.

src/Trustline.ts
dakingha69: Added a description

src/Trustline.ts
dakingha69: also added

PR #160, opened 2018-12-07
Only use yarn

### LINKED ISSUE #159: Support only one package manager
# Issue description
Currently we use `npm` and `yarn` at the same time. So we have two lock files which can a bit confusing and it is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files.

# Solution
Only use one package manager. Remove one lock file and change documentation and travis script accordingly.

dakingha69:
Closes #159


package.json
```diff
@@ -81,7 +81,7 @@
   },
   "dependencies": {
     "bignumber.js": "7.2.1",
-    "eth-lightwallet": "^3.0.1",
+    "eth-lightwallet": "dakingha69/eth-lightwallet",
```
berndbohmeier: whats this?

berndbohmeier: In this cases please use a fork under our name. This temporary things usually last longer and then only you can make changes. 
https://github.com/trustlines-network/eth-lightwallet

berndbohmeier: Just apply your commit

PR #162, opened 2018-12-10
adapt end2end tests to 'no fee for last hop' changes

schmir:
The fees for the last hop are now zero.

This makes the end2end test run with the fix/adapt-to-contracts-0.4 relay server branch


berndbohmeier:
@schmir This can be merged right?

schmir:
Apparently I didn't dare to merge it into the clientlib. Sorry. Will do now.

PR #165, opened 2018-12-11
Increase unit test coverage for class Configuration

dakingha69:
Add unit tests for the class `Configuration`. Also includes a small fix for `configuration.apiUrl`. The URL was created with an ending `/` which is removed in this PR.


dakingha69:
Moved into https://github.com/trustlines-network/clientlib/pull/166

PR #170, opened 2019-01-15
Fix eth-index service in docker compose

dakingha69:
Fixes docker-compose for running the e2e tests

### DIFF
--- a/tests/testrelay/docker-compose.yml
+++ b/tests/testrelay/docker-compose.yml
@@ -36,9 +36,9 @@ services:
       - PGDATABASE
       - PGPASSWORD
     restart: unless-stopped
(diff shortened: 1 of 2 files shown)


tests/testrelay/docker-compose.yml
```diff
@@ -36,14 +36,30 @@ services:
       - PGDATABASE
       - PGPASSWORD
     restart: unless-stopped
-    command: bash -c "sleep 15 && /opt/ethindex/bin/ethindex runsync --jsonrpc http://node:8545 --waittime 200"
+    command: runsync --jsonrpc http://node:8545 --waittime 200
```
berndbohmeier: Are you sure that you will not need the sleep?

dakingha69: It works without sleep because of the `restart: unless-stopped`.

berndbohmeier: okay

PR #176, opened 2019-01-21
Fix/use end2end repo 173

schmir:
use end2end repository to run end2end tests.

We already use the end2end repository to run those tests in the relay server, py-eth-index and contracts repository. With this commit we also do it for the clientlib repository.

If you feel the need to be able to run the end2end tests locally with yarn, we could also implement that in the end2end repo.


schmir:
> But yes it would definitely be nice to be able to run e2e tests locally with yarn.

ok. I'll look into it.

schmir:
I've added the functionality to run the end2end tests locally. Please take a look at the end2end repo's README: https://github.com/trustlines-network/end2end

It always tests the full set of tests. Let me know if that suffices.

dakingha69:
Ok cool thanks! Just checked it out, it should suffice for now :+1:

PR #180, opened 2019-01-23
Fix circle ci persist and attach paths

### LINKED ISSUE #177: CircleCI job deploy-dev fails
Publishing the `dev` version of the clientlib to npm fails because the package is too big as it also includes the docker image. See here https://circleci.com/gh/trustlines-network/clientlib/287.

dakingha69:
Closes #177

This should do the job hopefully.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
           name: Run Linter
           command: yarn lint
       - persist_to_workspace:
-          root: ~
+          root: "~"
           paths: .
 
   build:
@@ -56,14 +56,14 @@ jobs:
           name: Build Production
           command: yarn build
       - persist_to_workspace:
-          root: ~
+          root: "~"
           paths: .
 
   unit-test:
     <<: *defaults
     steps:
       - attach_workspace:
-          at: ~
+          at: "~"
       - run:
           name: Run Unit Tests
           command: yarn test:unit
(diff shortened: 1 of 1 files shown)


schmir:
while I'm here: I also think we should not persist in the `lint` job.
 It persists the same files that `build` persists and we should only persist additive changes (i.e. new files).

There's nothing to persist anyway after the linting process has been run, right?

schmir:
You could also attach the workspace and remove yarn-install in the lint job. That should speedup things a bit. You must make the lint job depend on build so that the yarn-install step is done.

berndbohmeier:
You write the npm token to ./repo/.npmrc and not to ./.npmrc, is this intended?
This risks uploading the token to npm, no?

dakingha69:
@berndbohmeier Yes that is intended. `npm publish` looks for a `.npmrc` file in the root folder of the repository, so it has to be under `~/repo`. There is also nothing to worry about because `npm publish` ignores files like `.npmrc`.

> npm will always ignore these file name patterns:
>     .\*.swp
>     ._*
>     .DS_Store
>     .git, hg , .svn, CVS version control directories
>     .npmrc
>     .lock-wscript
>     .wafpickle-*
>     config.gypi
>     npm-debug.log

https://blog.npmjs.org/post/165769683050/publishing-what-you-mean-to-publish

Issue #181, opened 2019-01-29
Implement RelayProvider and RelaySigner


dakingha69:
In order to migrate to `ethers.js`, we need to implement custom classes `RelayProvider` and `RelaySigner`, which contain all `eth-lightwallet` related functionalities. They should implement the interfaces provided by `ethers.js`(https://github.com/ethers-io/ethers.js/blob/master/providers/abstract-provider.d.ts)

Related to https://github.com/trustlines-network/project/issues/332

PR #184, opened 2019-01-30
Decouple Utils

dakingha69:
This PR removes the `Configuration` class and decouples `Utils` from its previous dependency on `Configuration`. While working on implementing the `RelayProvider` for issue #181, it became inconvenient that `Utils` had stored the `relayApiUrl` in its state as it would have resulted in a circular dependency.


src/Exchange.ts
```diff
@@ -547,7 +549,7 @@ export class Exchange {
    * @param payload Body of POST request.
    */
   private _postRequest(path: string, payload: any): Promise<any> {
-    return this.utils.fetchUrl(path, {
+    return this.provider.fetchEndpoint(path, {
```
berndbohmeier: I would make this an own method of the provider to
`postEndpoint(endpoint, payload)`

Does not have to be in this PR but if you agree open an issue.

tests/helpers/FakeTLProvider.ts
```diff
@@ -94,27 +95,16 @@ export class FakeUtils extends Utils {
       default:
         break
     }
+    console.log(response)
```
berndbohmeier: Do we need this log statement?

src/Contact.ts
```diff
@@ -1,18 +1,28 @@
...
+      this.user.address
+    }/contacts`
+    return this.provider.fetchEndpoint(endpoint)
   }
 
   public createLink(address: string, username: string): Promise<string> {
     return new Promise((resolve, reject) => {
       const params = ['contact', address, username]
-      resolve(this.utils.createLink(params))
+      resolve(createLink(params))
```
berndbohmeier: It is now hard to spot, that createLink is not defined here. Maybe it would be better to always import the utils, so that you have to write
`utils.createLink()`

src/providers/RelayProvider.ts
```diff
@@ -0,0 +1,36 @@
...
+export class RelayProvider implements TLProvider {
+  public relayApiUrl: string
+  public relayWsApiUrl: string
+
+  constructor(relayApiUrl: string, relayWsApiUrl: string) {
+    this.relayApiUrl = relayApiUrl
+    this.relayWsApiUrl = relayWsApiUrl
+  }
+
+  public async fetchEndpoint<T>(
```
berndbohmeier: What do you think of `fetch(endpoint, options)`
Do you think this has the same amount of information, or do you think `fetchEndpoint(endpoint, options)` is better?

dakingha69: I also thought about naming it only `fetch` but could also be confusing as there is already a fetch api in javascript :thinking: Or maybe `fetchJson`? I am pretty indifferent here..

src/Contact.ts
dakingha69: ok true. will change this

src/Exchange.ts
dakingha69: yes will open an issue

src/providers/RelayProvider.ts
dakingha69: I will leave it as `fetchEndpoint` for now.

PR #186, opened 2019-02-06
Feature/ethers relay provider signer

### LINKED ISSUE #181: Implement RelayProvider and RelaySigner

dakingha69:
Closes #181 

Sorry for big PR... But a lot of added lines are only mocks and fixtures for some tests. Most interesting stuff is in `src/providers/RelayProvider.ts` and `src/signers/EthersWallet.ts`.


src/providers/RelayProvider.ts
```diff
@@ -33,4 +60,341 @@ export class RelayProvider implements TLProvider {
...
+   * Returns a `Promise` with the latest block number.
+   */
+  public async getBlockNumber(): Promise<number> {
+    return this.fetchEndpoint<number>(`blocknumber`)
+  }
+
+  /**
+   * Returns a `Promise` with the current gas price as a `BigNumber`.
+   */
+  public async getGasPrice(): Promise<ethers.utils.BigNumber> {
```
berndbohmeier: Why did you not implement this? See line 112

src/signers/RelaySigner.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  constructor(provider: TLProvider) {
+    this.provider = provider
+  }
+
+  ///////////////
+  // Accessors //
+  ///////////////
+
+  public get address(): string {
```
berndbohmeier: We have now address and getAddress, I think we should only have one.

src/signers/RelaySigner.ts
```diff
@@ -0,0 +1,202 @@
...
+
+import { TLProvider } from '../providers/TLProvider'
+import { TLSigner } from './TLSigner'
+
+import { Signature, UserObject } from '../typings'
+
+/**
+ * The RelaySigner class contains wallet related methods.
+ */
+export class RelaySigner implements TLSigner {
```
berndbohmeier: Why is this called relay signer, what is relay specific.

src/providers/RelayProvider.ts
```diff
@@ -1,9 +1,13 @@
...
+import { ethers } from 'ethers'
 import { Observable } from 'rxjs/Observable'
 
 import { TLProvider } from './TLProvider'
 
 import utils from '../utils'
 
+import { TxInfos, TxInfosRaw } from '../typings'
+
 export class RelayProvider implements TLProvider {
```
berndbohmeier: Why does this have to implement both and not just the ethers provider. I think we might mix concepts here, lets have a call tomorrow.

src/providers/RelayProvider.ts
```diff
@@ -33,4 +60,341 @@ export class RelayProvider implements TLProvider {
...
+   * @param signedTransaction
+   */
+  public async sendTransaction(
+    signedTransaction: string | Promise<string>
+  ): Promise<ethers.providers.TransactionResponse> {
+    if (typeof signedTransaction !== 'string') {
+      signedTransaction = await signedTransaction
+    }
+    const transaction = ethers.utils.parseTransaction(signedTransaction)
+    await this.relayTx(signedTransaction)
```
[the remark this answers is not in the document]
dakingha69: Ok opened an issue https://github.com/trustlines-network/clientlib/issues/187

src/providers/RelayProvider.ts
dakingha69: The problem is that our `/users/:userAddress/txinfos` endpoint requires the address of the user. The method `getGasPrice()` on the other hand does not allow any parameters. I guess I can use the ZERO_ADDRESS then instead? Smth like
```javascript
return this.fetchEndpoint('/users/0x0...0/txinfos')
```

src/signers/RelaySigner.ts
dakingha69: Yes they are a bit redundant. I also wanted to only have `address` but `getAddres()` is a property from `ethers.Signer`. So maybe we should only keep `getAddress()`?

src/signers/RelaySigner.ts
dakingha69: Hmm yes nothing truly relay specific related here... Maybe rename it to `WalletSigner`, `Wallet` or `TLWallet`?

src/providers/RelayProvider.ts
dakingha69: I wasn't 100% sure about this. I thought that if we want to support other providers like `web3`. `etherscan`, ... it would be necessary to have another abstraction `TLProvider`. But let us talk about this in the call.

src/signers/RelaySigner.ts
berndbohmeier: Lets take TL Wallet

src/TLNetwork.ts
```diff
@@ -80,7 +80,7 @@ export class TLNetwork {
   /**
    * @hidden
    */
-  public signer: TxSigner
+  public signer: TLWallet
```
dakingha69: Not sure if this should still be called signer or change it to wallet?

src/User.ts
```diff
@@ -13,13 +13,15 @@ import { Amount, RawTxObject, Signature, UserObject } from './typings'
  * related methods.
  */
 export class User {
-  private signer: TxSigner
+  private signer: TLWallet
   private transaction: Transaction
   private provider: TLProvider
 
+  private defaultPassword = 'ts'
```
dakingha69: I am following the implementation we did with eth-lightwallet here. Do we want to make the password mandatory or optional? It is mandatory in ethers.js

src/signers/Web3Wallet.ts
```diff
@@ -1,13 +1,13 @@
...
 
-import { TxSigner } from './TxSigner'
+import { TLWallet } from './TLWallet'
 
 import { Amount, RawTxObject, Signature, TxInfos, UserObject } from '../typings'
 
 /**
  * The Web3Signer class contains functions for signing transactions with a web3 provider.
  */
-export class Web3Signer implements TxSigner {
```
berndbohmeier: Do we not have to change the implementation of `confirm` here?

src/Transaction.ts
```diff
@@ -14,10 +15,10 @@ const ETH_DECIMALS = 18
  * The Transaction class contains functions that are needed for Ethereum transactions.
  */
 export class Transaction {
-  private signer: TxSigner
+  private signer: TLWallet
```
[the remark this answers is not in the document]
berndbohmeier: signer interface would be similar to https://github.com/ethers-io/ethers.js/blob/master/src.ts/abstract-signer.ts

src/TLNetwork.ts
berndbohmeier: After our discussion now, this should probably stay signer.

src/User.ts
dakingha69: https://github.com/trustlines-network/clientlib/issues/189

src/User.ts
```diff
@@ -13,50 +14,67 @@ import { Amount, RawTxObject, Signature, UserObject } from './typings'
...
-    this.provider = params.provider
+    this.wallet = params.wallet
   }
 
   /**
    * Checksummed Ethereum address of currently loaded user/keystore.
    */
   public get address(): string {
-    return this.signer.address
+    return this.wallet.address
```
berndbohmeier: I think this should be signer.address, no?

berndbohmeier: Or lets create an issue to solve that another time.

dakingha69: https://github.com/trustlines-network/clientlib/issues/190

PR #191, opened 2019-02-11
Remove eth-lightwallet dependency

### LINKED ISSUE #182: Remove eth-lightwallet dependency
As discussed here #164, we are migrating to `ethers.js` and therefore the dependency on `eth-lightwallet` should be removed.

Depends on #181 and should adapt library to use new implemented provider and signer.

dakingha69:
Closes #182 

Removal of library and all artefacts related to eth-lightwallet.


tests/browser/index.html
```diff
@@ -1,6 +1,7 @@
 <html>
   <head>
-    <script type="text/javascript" src="../../_bundles/trustlines-network.js"></script>
+    <script src="https://cdn.ethers.io/scripts/ethers-v4.min.js" charset="utf-8" type="text/javascript"></script>
```
berndbohmeier: I dont think it makes sense to have ethers external, if it is needed to work. 
Is this the same bundle we use in the app?

dakingha69: So in general we create 3 different formats which support different approaches to import our library.
```javascript
// ./lib-esm -> ES6 
import { TLNetwork } from 'trustlines-clientlib'

// lib -> UMD format
const TLNetwork = require('webpack-numbers').TLNetwork
 
// _bundles -> loading lib via script tag
<script type="text/javascript" src="../../_bundles/trustlines-clientlib.js"></script>
```
The first 2 formats are created via `tsc` whereas the last one is the only one bundled with `webpack`. And the app uses `lib-esm`.

dakingha69: https://github.com/trustlines-network/clientlib/issues/192

PR #193, opened 2019-02-13
Remove web3

### LINKED ISSUE #183: Remove web3 dependency
As discussed here #164, we are migrating to `ethers.js` and therefore the dependency on `web3` should be removed.

Depends on #181 and should adapt library to use new implemented provider and signer.

dakingha69:
Closes #183 

- removes web3 dependency and artifacts
- implements missing methods in `Web3Signer` with ethers

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,13 +82,11 @@
   },
   "dependencies": {
     "bignumber.js": "7.2.1",
-    "ethereumjs-abi": "^0.6.5",
     "ethereumjs-util": "^5.2.0",
     "ethers": "^4.0.23",
     "html5-websocket": "^2.0.2",
     "rxjs": "^5.4.1",
     "simple-jsonrpc-js": "0.0.10",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -82,13 +82,11 @@
   },
   "dependencies": {
     "bignumber.js": "7.2.1",
-    "ethereumjs-abi": "^0.6.5",
     "ethereumjs-util": "^5.2.0",
```
berndbohmeier: Do we still need ethereumjs-utils?

dakingha69: Yes we use some ethereumjs-utils functionalities here and there. Will open an issue to replace this also with `ethers.js`.

dakingha69: https://github.com/trustlines-network/clientlib/issues/199

dakingha69:
Only tested it locally with a button that triggers the signing. Opened an issue for this https://github.com/trustlines-network/clientlib/issues/200.

PR #202, opened 2019-03-05
Feature/receiver pays path

cducrest:
Add an option to get a path and make a payment for receiver pays the fees
Also return fee payer for closing path function.

closes: https://github.com/trustlines-network/clientlib/issues/197

### DIFF
--- a/src/Payment.ts
+++ b/src/Payment.ts
@@ -11,6 +11,7 @@ import utils from './utils'
 import {
   EventFilterOptions,
   FeePayer,
+  isFeePayerValue,
   NetworkTransferEvent,
   PathObject,
   PathRaw,
@@ -188,6 +189,11 @@ export class Payment {
       headers: new Headers({ 'Content-Type': 'application/json' }),
       method: 'POST'
     })
+
+    if (!isFeePayerValue(feePayer)) {
+      throw Error(`Unexpected feePayer value: ${feePayer}`)
+    }
+
     return {
       estimatedGas: new BigNumber(estimatedGas),
       feePayer: feePayer as FeePayer,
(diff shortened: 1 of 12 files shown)


src/Payment.ts
```diff
@@ -157,27 +160,37 @@ export class Payment {
...
+      path,
+      feePayer
+    } = await this.provider.fetchEndpoint<PathRaw>(endpoint, {
       body: JSON.stringify(data),
       headers: new Headers({ 'Content-Type': 'application/json' }),
       method: 'POST'
     })
     return {
       estimatedGas: new BigNumber(estimatedGas),
+      feePayer: feePayer as FeePayer,
```
berndbohmeier: I think we should check this to make sure this is true. Otherwise types will not help us. 
It should be possible with something like:
`if Object.values(FeePayer).includes(feePayer)`

src/Trustline.ts
```diff
@@ -383,6 +385,7 @@ export class Trustline {
 
     return {
       estimatedGas: new BigNumber(estimatedGas),
+      feePayer: feePayer as FeePayer,
```
berndbohmeier: same as above

tests/unit/Typings.test.ts
```diff
@@ -0,0 +1,26 @@
+import BigNumber from 'bignumber.js'
+import * as chai from 'chai'
+import * as chaiAsPromised from 'chai-as-promised'
+import fetchMock = require('fetch-mock')
+import 'mocha'
+
+import utils from '../../src/utils'
+
+import {
+  ExchangeCancelEvent,
```
berndbohmeier: What is this? Can you please clean that up?

PR #203, opened 2019-03-05
Add a check with prettier in circleci

cducrest:
Add prettier check to circleci


cducrest:
Sorry, I must have been too tired at same time, should be good now.

PR #206, opened 2019-03-12
Feature/identity wallet

cducrest:
Add the account management part of the identity wallet, responsible for asking the relay server to deploy an identity contract and saving / restoring a wallet

I am unsure as whether I need to modify https://github.com/trustlines-network/clientlib/blob/develop/src/TLNetwork.ts#L128 to add a possibility for a user with an IdentityWallet. I don't know what the purpose of this class is.

https://github.com/trustlines-network/clientlib/issues/196

### DIFF
--- a/src/Exchange.ts
+++ b/src/Exchange.ts
@@ -550,7 +550,7 @@ export class Exchange {
    * @param payload Body of POST request.
    */
   private _postRequest(path: string, payload: any): Promise<any> {
-    return this.provider.PostToEndpoint(path, payload)
+    return this.provider.postToEndpoint(path, payload)
   }
 
   /**
--- a/src/Exchange.ts
+++ b/src/Exchange.ts
@@ -550,11 +550,7 @@ export class Exchange {
    * @param payload Body of POST request.
    */
   private _postRequest(path: string, payload: any): Promise<any> {
-    return this.provider.fetchEndpoint(path, {
-      body: JSON.stringify(payload),
-      headers: new Headers({ 'Content-Type': 'application/json' }),
-      method: 'POST'
-    })
+    return this.provider.PostToEndpoint(path, payload)
   }
 
   /**
--- a/src/User.ts
+++ b/src/User.ts
@@ -10,7 +10,7 @@ import utils from './utils'
 import { Amount, RawTxObject, Signature, UserObject } from './typings'
 
 /**
- * The User class contains all user related functions, which also include backup
+ * The User class contains all user related functions, which also include wallet
  * related methods.
  */
 export class User {
@@ -34,21 +34,21 @@ export class User {
   }
 
   /**
-   * Checksummed Ethereum address of currently loaded user/backup.
+   * Checksummed Ethereum address of currently loaded user/wallet.
    */
   public get address(): string {
     return this.wallet.address
   }
 
   /**
-   * Public key of currently loaded user/backup.
+   * Public key of currently loaded user/wallet.
    */
   public get pubKey(): string {
     return this.wallet.pubKey
   }
 
   /**
-   * Creates a new user and the respective backup using the configured signer.
+   * Creates a new user and the respective wallet using the configured signer.
    * Loads new user into the state and returns the created user object.
    * @param progressCallback Optional progress callback to call on encryption progress.
    */
@@ -61,17 +61,17 @@ export class User {
   }
 
   /**
-   * Loads an existing user and respective backup.
+   * Loads an existing user and respective wallet.
    * Returns the loaded user object.
-   * @param backup The string backup gotten from createAccount() of a TLWallet.
+   * @param serializedWallet The string serialized wallet gotten from createAccount() of a TLWallet.
    * @param progressCallback Optional progress callback to call on encryption progress.
    */
   public async load(
-    backup: string,
+    serializedWallet: string,
     progressCallback?: any
   ): Promise<UserObject> {
     const loadedAccount = await this.wallet.loadAccount(
-      backup,
+      serializedWallet,
       this.defaultPassword,
       progressCallback
     )
@@ -79,7 +79,7 @@ export class User {
   }
 
   /**
-   * Digitally signs a message hash with the currently loaded user/backup.
+   * Digitally signs a message hash with the currently loaded user/wallet.
    * @param msgHash Hash of message that should be signed.
    */
   public async signMsgHash(msgHash: string): Promise<Signature> {
@@ -126,7 +126,7 @@ export class User {
   }
 
   /**
-   * Recovers user / backup from 12 word seed.
+   * Recovers user / wallet from 12 word seed.
    * @param seed 12 word seed phrase string.
    * @param progressCallback Optional progress callback to call on encryption progress.
    */
@@ -147,17 +147,16 @@ export class User {
    * and are willing to send some of it to the new user. The function is called by a
    * new user who wants to get onboarded, respectively has no ETH or trustline.
    * @param username Name of new user who wants to get onboarded.
-   * @param serializedKeystore Serialized [eth-lightwallet](https://github.com/ConsenSys/eth-lightwallet)
-   *                           backup of new user who wants to get onboarded.
+   * @param serializedWallet Serialized wallet of new user who wants to get onboarded.
    * @param progressCallback Optional progress callback to call on encryption progress.
    */
   public async createOnboardingMsg(
     username: string,
-    serializedKeystore: string,
+    serializedWallet: string,
     progressCallback?: any
   ): Promise<string> {
     const { address, pubKey } = await this.wallet.loadAccount(
-      serializedKeystore,
+      serializedWallet,
       this.defaultPassword,
       progressCallback
     )
--- a/src/User.ts
+++ b/src/User.ts
@@ -64,15 +64,18 @@ export class User {
    * Loads an existing user and respective keystore.
    * Returns the loaded user object.
    * @param serializedKeystore Serialized [eth-lightwallet](https://github.com/ConsenSys/eth-lightwallet) key store.
+   * @param identityAddress Optional the address of the identity contract to load in case of Identity Wallet
    * @param progressCallback Optional progress callback to call on encryption progress.
    */
   public async load(
     serializedKeystore: string,
+    identityAddress?: string,
     progressCallback?: any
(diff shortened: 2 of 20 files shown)


src/providers/RelayProvider.ts
```diff
@@ -30,6 +30,15 @@ export class RelayProvider implements TLProvider {
     return utils.fetchUrl<T>(`${this.relayApiUrl}/${trimmedEndpoint}`, options)
   }
 
+  public async fetchPostEndpoint<T>(endpoint: string, data: any): Promise<T> {
```
berndbohmeier: I would just call it `postEndpoint` or `postToEnpoint`. I think the fetch is redundant.

src/typings.ts
```diff
@@ -359,6 +359,18 @@ export interface UserObject {
...
+export interface IdentityUserObject {
+  identityAddress: string
+  pubKey: string
+  keystore: string
+}
+
+export interface DeployedIdentity {
+  identity: string
+  nextNonce: number
+  blance: string
```
berndbohmeier: balance

src/wallets/IdentityWallet.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  constructor(provider: TLProvider) {
+    this.provider = provider
+  }
+
+  public get address(): string {
+    return this.identityAddress
+  }
+
+  public get pubKey(): string {
```
berndbohmeier: This is not the right pub key. We can not implement that right now, so just throw a not implemented error.

src/wallets/IdentityWallet.ts
```diff
@@ -0,0 +1,179 @@
...
+   * Decrypts given keystore and loads wallet.
+   * @param encryptedKeystore Encrypted keystore from `createAccount`.
+   * @param password Password to decrypt keystore.
+   * @param identityAddress the address of the corresponding identity contract
+   * @param progressCallback Callback function for decryption progress.
+   */
+  public async loadAccount(
+    encryptedKeystore: string,
+    password: string,
+    identityAddress: string,
```
berndbohmeier: Here you change the interface. Please dont do that, as this complicates using this library. 
Just try to store the address somehow in the keystore. It does not have to be encrypted I think.

src/wallets/IdentityWallet.ts
```diff
@@ -0,0 +1,179 @@
...
+   * Recovers wallet from mnemonic phrase and encrypts keystore with given password.
+   * @param seed Mnemonic seed phrase.
+   * @param password Password to encrypt recovered keystore.
+   * @param identityAddress the address of the corresponding identity contract
+   * @param progressCallback Callback function for encryption progress.
+   */
+  public async recoverFromSeed(
+    seed: string,
+    password: string,
+    identityAddress: string,
```
berndbohmeier: I also dont  know how that makes sense, the user will only have backuped the seed, so this does not help. We should maybe discuss this.

src/wallets/TLWallet.ts
```diff
@@ -12,11 +12,13 @@ export interface TLWallet {
   loadAccount(
     serializedKeystore: string,
     password: string,
+    identityAddress?: string,
```
berndbohmeier: Dont change the interface

src/providers/RelayProvider.ts
cducrest: Sounds good

cducrest:
Should be good.
I no longer change the interface, and use a `backup` with versions, etc ... instead of a single `encryptedKeystore`.
The old `encryptedKeystore`s should still be usable to load a wallet.
I kept `recoverFromPrivateKey()` in identity wallet with the address as parameter as it is not in the `TLWallet` interface.

The only thing I am unsure about is if I could have avoided hardcoding the backup in the fixture for testing: https://github.com/trustlines-network/clientlib/pull/206/files#diff-805eca9615f90383b5b0c55dfe305144R76

cducrest:
As `IdentityWallet` is not a `TLSigner`, it would take too many modifications (or make no sense) to handle it within `TLNetwork.ts`.

src/typings.ts
```diff
@@ -356,14 +360,33 @@ export interface DecimalsObject {
 export interface UserObject {
   address: string
   pubKey: string
-  keystore: string
+  backup: string
+}
+
+export interface DeployedIdentity {
+  identity: string
```
berndbohmeier: is this the address? then we should call it address. It does not make much sense that a structure Identity has a identity.

src/wallets/EthersWallet.ts
```diff
@@ -60,40 +66,86 @@ export class EthersWallet implements TLWallet, TLSigner {
...
 
   /**
-   * Encrypts given keystore and loads wallet.
-   * @param encryptedKeystore Encrypted keystore from `createAccount`.
-   * @param password Password to decrypt keystore.
+   * Create a backup following the current backup format from an encryptedKeystore
+   * Can be used to migrate from backup version 0 to backup version 1
+   * @param encryptedKeystore
+   */
+  public createBackup(encryptedKeystore: string): string {
```
berndbohmeier: `serializeWallet` or similar. It should describe that it just creates a serialized version of the wallet and returns that. There is no backup done.

src/typings.ts
```diff
@@ -356,14 +360,33 @@ export interface DecimalsObject {
...
+  nextNonce: number
+  balance: string
 }
 
 export interface Signature {
   ecSignature: ECSignature
   concatSig: string
 }
 
+export interface IdentityBackup {
```
berndbohmeier: It is not a backup, it is just a format that can be used to store the wallet for any reason.

tests/e2e/Identity.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  const relayProvider = new RelayProvider(relayApiUrl, relayWpUrl)
+
+  const DEFAULT_PASSWORD = 'ts'
+  const ACCOUNT_KEYS = ['address', 'backup', 'pubKey']
+
+  describe('Deploy identity by creating an identity account', async () => {
+    const identityWallet = new IdentityWallet(relayProvider)
+    const createdAccount = await identityWallet.createAccount(DEFAULT_PASSWORD)
+    assert.hasAllKeys(createdAccount, ACCOUNT_KEYS)
+    expect(createdAccount.address).to.not.equal(undefined)
```
berndbohmeier: can you maybe at least check that it is a non empty string or better it is a valid address? Like this it could forexample just return null and we would not see that.

src/typings.ts
```diff
@@ -356,14 +360,33 @@ export interface DecimalsObject {
 export interface UserObject {
   address: string
   pubKey: string
-  keystore: string
+  backup: string
```
berndbohmeier: This breaks backwards compatibility, do we really need that?
If we break like this we should make it clear in the PR or Commit message.

src/wallets/EthersWallet.ts
```diff
@@ -60,40 +66,86 @@ export class EthersWallet implements TLWallet, TLSigner {
       password,
       typeof progressCallback === 'function' && progressCallback
     )
+
+    const backup = this.createBackup(encryptedKeystore)
+
     return {
       address: this.address,
-      keystore: encryptedKeystore,
+      backup,
```
berndbohmeier: see above, breaks backwards compatibility.

src/Exchange.ts
```diff
@@ -550,11 +550,7 @@ export class Exchange {
    * @param payload Body of POST request.
    */
   private _postRequest(path: string, payload: any): Promise<any> {
-    return this.provider.fetchEndpoint(path, {
-      body: JSON.stringify(payload),
-      headers: new Headers({ 'Content-Type': 'application/json' }),
-      method: 'POST'
-    })
+    return this.provider.PostToEndpoint(path, payload)
```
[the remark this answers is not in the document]
cducrest: This is already outdated

src/typings.ts
cducrest: I just had it match the fields of the relay server API so I can do things like:
`const identity = await this.provider.postToEndpoint<DeployedIdentity>`
But I will change it to identityAddress.

cducrest: Actually, the only alternative solution I found is to use `postToEndpoint<any>()` which I believe is worse. So i changed the type name to `DeployIdentityResponse` and kept `identity` as a field.

src/User.ts
```diff
@@ -61,25 +61,25 @@ export class User {
   }
 
   /**
-   * Loads an existing user and respective keystore.
+   * Loads an existing user and respective wallet.
    * Returns the loaded user object.
-   * @param serializedKeystore Serialized [eth-lightwallet](https://github.com/ConsenSys/eth-lightwallet) key store.
+   * @param serialisedWallet The string serialised wallet gotten from createAccount() of a TLWallet.
```
berndbohmeier: serialize. In programming you use american english usually, like color instead of colour.

src/User.ts
```diff
@@ -148,7 +148,7 @@ export class User {
    * new user who wants to get onboarded, respectively has no ETH or trustline.
    * @param username Name of new user who wants to get onboarded.
    * @param serializedKeystore Serialized [eth-lightwallet](https://github.com/ConsenSys/eth-lightwallet)
```
berndbohmeier: why did you not change this here? this should also be serializedWallet now. It is handed over to load

berndbohmeier: Also the eth-lightwallet comment is totally outdated, please remove that.

src/typings.ts
```diff
@@ -374,14 +375,14 @@ export interface Signature {
   concatSig: string
 }
 
-export interface IdentityBackup {
+export interface DeserialisedIdentityWallet {
```
berndbohmeier: That does not make sense. I dont know what a deserializedWallet should be.

berndbohmeier: Maybe `IdentityWalletSerializationFormat` ?

cducrest: I don't know what this means.

src/User.ts
cducrest: Because I struggled to understand what this function was used for, but afaict it is only useful for an ethers wallet and not for an identity wallet. The serializedKeystore will work for an ethers wallet so I did not want to change the function because the way it is used in the mobile app would need to change accordingly.
I can do it if you wish.

PR #208, opened 2019-03-21
Add method to sign and confirm a meta-transaction to identity wallet

cducrest:
closes: https://github.com/trustlines-network/clientlib/issues/194

### DIFF
--- a/src/typings.ts
+++ b/src/typings.ts
@@ -208,6 +208,16 @@ export interface RawTxObject {
   nonce?: number
 }
 
+export interface MetaTransaction {
+  extraData: string
+  from: string
+  to: string
+  value: string
+  data: string
+  nonce: string
+  signature?: string
+}
+
 export interface Web3TxReceipt {
   status: boolean
   blockHash: string
--- a/src/wallets/IdentityWallet.ts
+++ b/src/wallets/IdentityWallet.ts
@@ -4,14 +4,20 @@ import { TLProvider } from '../providers/TLProvider'
 import { TL_WALLET_VERSION, TLWallet, WALLET_TYPE_IDENTITY } from './TLWallet'
 
 import {
+  Amount,
   DeployIdentityResponse,
   IdentityWalletSchema,
+  MetaTransaction,
+  RawTxObject,
+  Signature,
   UserObject
 } from '../typings'
 
-export class IdentityWallet implements TLWallet {
-  // TODO: make this class a TLSigner as part of https://github.com/trustlines-network/clientlib/issues/194
+import { joinSignature, SigningKey } from 'ethers/utils'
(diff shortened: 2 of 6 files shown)


src/wallets/IdentityWallet.ts
```diff
@@ -225,4 +317,27 @@ export class IdentityWallet implements TLWallet {
...
+      )
+    }
+  }
+
+  private buildMetaTransaction(rawTx: RawTxObject): MetaTransaction {
+    const metaTransaction: MetaTransaction = {
+      data: ethers.utils.solidityKeccak256(['bytes'], [rawTx.data]),
+      extraData: '0x',
+      from: rawTx.from,
+      nonce: rawTx.nonce.toString(),
```
berndbohmeier: why does that have to be a string?

src/typings.ts
```diff
@@ -208,6 +208,16 @@ export interface RawTxObject {
   nonce?: number
 }
 
+export interface MetaTransaction {
+  extraData: string
+  from: string
+  to: string
+  value: string
+  data: string
+  nonce: string
```
berndbohmeier: why is this a string?

src/wallets/IdentityWallet.ts
```diff
@@ -186,6 +202,82 @@ export class IdentityWallet implements TLWallet {
...
+    ]
+    const values = [
+      '0x19',
+      '0x00',
+      metaTransaction.from,
+      metaTransaction.to,
+      metaTransaction.value,
+      ethers.utils.solidityKeccak256(['bytes'], [metaTransaction.data]),
+      metaTransaction.nonce
+    ]
```
berndbohmeier: You forgot the extraData

tests/e2e/Identity.test.ts
```diff
@@ -29,18 +33,49 @@ describe('e2e', () => {
...
+      it('should relay meta transaction and return a transaction hash', async () => {
+        const rawTx: RawTxObject = {
+          data: '0x',
+          from: identityWallet.address,
+          nonce: 1,
+          to: identityWallet.address,
+          value: 0
+        }
+
+        const transactionHash = await identityWallet.confirm(rawTx)
```
berndbohmeier: This is the transactionHash of the ethereum transaction? Is there a way to get the hash of the metaTransaction?

src/wallets/IdentityWallet.ts
cducrest: I can see in the relay server that it is interpreted in the `MetaTransactionSchema` as the same type as `value` (BigInteger) which afaict should be a string.

berndbohmeier: Ah I am always confused that some of these types are types of the api. I'll create an issue to make that explicit.

PR #210, opened 2019-03-26
Feature/meta transaction provider methods

cducrest:
Should finish the identity story for the clientlib.
Does a little more than https://github.com/trustlines-network/clientlib/issues/195 because of refactoring.

All e2e tests are parametrized so that they run both using ethers wallets and identity wallets except for `Exchange.test.ts` because it does not support identity and `Event.test.ts` because it uses `Exchange`. I could reformat it so that part of the `Event` tests run with identity wallets.
Since I had to change the tabulation of most e2e source files, the diff is made bigger than necessary.

I had an error message while building locally and on circleci, so I added a few lines that I do not understand in `webpack.config.js`: https://github.com/trustlines-network/clientlib/blob/feature/meta-transaction-provider-methods/webpack.config.js#L10

I had one random e2e test alternating between failing and succeeding locally with no change, but never on circle ci. I don't know what to do about it. (`Payment.test.ts` #prepareClose() with identity wallets and non-triangulated transfer, and `Trustline.test.ts` #getMaxAmountAndPathInNetwork for non-adjacent users with identity wallets)

I am not sure about https://github.com/trustlines-network/clientlib/blob/feature/meta-transaction-provider-methods/src/signers/Web3Signer.ts#L114 as I am not sure of the meaning of gasPrice returned by the relay server and if it is coherent with the web3Provider, but it should be correct.


README.md
```diff
@@ -31,13 +31,13 @@ Use the following configuration to connect to the currently deployed test setup.
 ```javascript
 import { TLNetwork } from 'trustlines-clientlib'
 
-const config = {
+const TLNetworkConfig = {
```
berndbohmeier: Why did you rename this? variable names should start with lower case letter

src/TLNetwork.ts
```diff
@@ -195,17 +182,34 @@ export class TLNetwork {
     this.provider = provider
   }
 
-  public setSigner(signer: TLSigner): void {
-    if (!(signer instanceof Web3Signer || signer instanceof EthersWallet)) {
+  public setSigner(web3Provider, wallet): void {
```
berndbohmeier: please use typings.
also I think it should be `signer` nor `wallet`

src/providers/RelayProvider.ts
```diff
@@ -102,4 +119,18 @@ export class RelayProvider implements TLProvider {
...
+   * Send the given signed meta-transaction to a relay server to execute it on the
+   * blockchain and returns a `Promise` with the transaction hash.
+   * @param metaTransaction Signed meta-transaction to be sent to the relay server
+   * @returns The hash of the transaction sent by the relay server, not to be confused with the hash of the meta-transaction
+   */
+  public async sendSignedMetaTransaction(
+    metaTransaction: MetaTransaction
+  ): Promise<string> {
+    return this.postToEndpoint<string>('relay-meta-transaction', {
+      metaTransaction
```
berndbohmeier: `signedMetaTransaction` ?

src/signers/Web3Signer.ts
```diff
@@ -98,4 +100,25 @@ export class Web3Signer implements TLSigner {
     const msgHashBytes = ethers.utils.arrayify(msgHash)
     return this.signMessage(msgHashBytes)
   }
+
+  public async getTxInfos(userAddress: string): Promise<TxInfos> {
```
berndbohmeier: How did you test this part of the code?

tests/Fixtures.ts
```diff
@@ -8,13 +9,26 @@ export const keystore2 = `{"encSeed":{"encStr":"UJrWA2ZaPF6lnWVktF/VjtLFJ3EZRpmh
 export const keystore3 =
   '{"encSeed":{"encStr":"eEM0n29iCxQIm67Xt5XQseVllcfNEX1PTV5COHBxaUlmyOASd8jmTRjB1ruItlUqB0+T758EQz1oCgRfW6oluGI2SEF5liaNf4ku2wqpgMMJX3xKhiCQ4oeSEAM/SfmZ/TQ4S+14crRgsZxS8fDdu2G8uOT6DWJ3LpOY3aKW34BIA5YepfjrFg==","nonce":"Wsc2oXdrXX/LSbZ4n6pciSHd+MPJ5eeN"},"encHdRootPriv":{"encStr":"qHQ6BXF6SZoku0++cPMfG3/cOmqR+Zzk95CCk0vnMUluvbIaqEK+wN4Sc0trcnUmIOoOtFrO15lgJETMcjB/nHBLAmSdeVH92rLBB5GIMA2AHXf4GYXETgl/Z8rq6Rm3GvqLFKwghewFspvgV5Ykjritd7/yajxsSKye7DcW1Q==","nonce":"9MfuGFodFQP9fmD3+OUz9oKCURDd16r/"},"addresses":["c7d6401b81a2f70c5906e8a6fbebe680c69fa03d"],"encPrivKeys":{"c7d6401b81a2f70c5906e8a6fbebe680c69fa03d":{"key":"nrum/qp+I6oF+AypB/6xVDdbmH7S4plVf3VxSCmRwiz8rF3HDyLwafYh3Hur4foC","nonce":"/HC96G1H5OCXCJSPN5Z5Rq4++3d1C+TO"}},"hdPathString":"m/44\'/60\'/0\'/0","salt":"pXfIg1o/mPRVUwiXBkYnl0WpJfu4oRqi61rjZx1SfNY=","hdIndex":1,"version":3}'
 
-export const config = {
+export const TLNetworkConfig = {
```
berndbohmeier: no capital letter for variable

README.md
cducrest: I did not find config to be explicit enough in most parts of the code and renamed it to TLNetworkConfig, I can understand that it does not matter in this particular place actually.
I can rename all of them to tlNetworkConfig if you want or back to config.

src/TLNetwork.ts
cducrest: If I use typing and signer, I will have: `public setSigner(web3Provider, signer: TLSigner)` and will not be able to pass the wallet (that is actually also a signer) to the function
If I use  `public setSigner(web3Provider, signer: TLWallet)`, I cannot do `this.signer = signer` because `this.signer` is of type `TLSigner`.

To solve the problem, since every wallet is a signer I would change the interface `TLWallet` to extend `TLSigner`. This way, we can pass a wallet to a function that expects a signer.

I wanted to do this as part of https://github.com/trustlines-network/clientlib/issues/205 but can do it here if you are find with this approach.

src/providers/RelayProvider.ts
cducrest: I can use `signedMetaTransaction` but the relay server API expects `"metaTransaction"`.
So it will result in `return this.postToEndpoint<string>('relay-meta-transaction', { metaTransaction: signedMetaTransaction })`.
Do you think it is better?

src/signers/Web3Signer.ts
cducrest: Not tested, I thought we'd have tests with the making a transaction with the Web3Signer already, and this would have been used by it.
I will try to add tests but am not sure how to create a web3signer object yet.

README.md
berndbohmeier: i am fine with tlNetworkConfig

src/TLNetwork.ts
berndbohmeier: if it is a signer, you should be able to pass it.

berndbohmeier: Ah, got it. No you can seperate that.

src/providers/RelayProvider.ts
berndbohmeier: how other parts are named should not limit us. Yes I think this is better.

src/signers/Web3Signer.ts
berndbohmeier: I think at the moment we can only manually test that. Please ask Dongha

cducrest: I added a unit test that made explicit missing parenthesis, in this method. I'll ask Dong-Ha for manual testing.

berndbohmeier:
Please make the manual test or ask Dongha to do it. Otherwise I think this is good now.

dakingha69:
> Please make the manual test or ask Dongha to do it. Otherwise I think this is good now.
I tested it manually and it works. I will also open a PR soon with a basic app which uses MetaMask and injected web3 so everyone can test Web3Signer functionality.

PR #211, opened 2019-03-27
Remove obsolete on-boarding message parameter

### LINKED ISSUE #207: What is createOnboardingMsg supposed to do
When adding the identity wallet, we ended up modifying createOnboardingMsg() :
https://github.com/trustlines-network/clientlib/blob/develop/src/User.ts#L153.
We changed its input from `ethersKeystore` to `serializedWallet` but were unsure of what the function is used for and if it needs to be modified / deleted.

My understandingis that it is used to create a message to ask for ETH to another user. It takes as input the ethersKeystore to be able to recover the address that should receive the eth. This is not necessary as it could use `this.address`.

weilbith:
Closes #207

### DIFF
--- a/src/User.ts
+++ b/src/User.ts
@@ -147,20 +147,9 @@ export class User {
    * and are willing to send some of it to the new user. The function is called by a
    * new user who wants to get onboarded, respectively has no ETH or trustline.
    * @param username Name of new user who wants to get onboarded.
-   * @param serializedWallet Serialized wallet of new user who wants to get onboarded.
-   * @param progressCallback Optional progress callback to call on encryption progress.
    */
-  public async createOnboardingMsg(
-    username: string,
-    serializedWallet: string,
-    progressCallback?: any
-  ): Promise<string> {
-    const { address, pubKey } = await this.wallet.loadAccount(
-      serializedWallet,
-      this.defaultPassword,
-      progressCallback
-    )
-    const params = ['onboardingrequest', username, address, pubKey]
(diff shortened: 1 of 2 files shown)


weilbith:
I'm a bit confused. Yes, one line will not executed anymore, since it has been removed. So the `Lines` as well as the `Hits` value has decreased. But I thought the coverage should be relative and not absolute.
I'm not sure what in [this metric](https://docs.codecov.io/docs/about-code-coverage#section-what-is-code-coverage) is the difference between `hits` and `sum of hit`. Probably this is a double notation of `sum of` and `+ ... +`.
> hits / (sum of hit + partial + miss)

This is for me this results into this:

|   | hits | sum of ... | coverage |
|:---:|:---:|:---:|:---:|
| develop | 589 | 892 | 0,66 |
| fix/... | 588 | 891 | 0,66 |

cducrest:
Your result is correct, I don't see what confuses you:
589/892 = 66.03
588/891 = 65.99
66.03 - 65.99 = 0.04

We lose 0.04% test coverage by removing one line that was previously covered.

Otherwise lgtm

berndbohmeier:
Maybe stop rounding?^^

weilbith:
Time for a new calculation tool. Too primitive it seems like.
I'm still confused why it is this way. I mean it makes sense from the math perspective. But in fact the coverage itself from the higher view hasn't changed. Anyway doesn't matter. Now I know that I have to check this value twice if I see the coverage has dropped.

PR #213, opened 2019-03-28
Run coverage on end2end tests

berndbohmeier:


berndbohmeier:
@schmir An idea why I get connection errors on circle?

schmir:
> @schmir An idea why I get connection errors on circle?

yes, all of the containers are being run on a different host on circle ci. you can choose the host the tests are using by setting the environment variable  `RELAY_HOSTNAME`.

According to the documentation at https://circleci.com/docs/2.0/building-docker-images/#accessing-services, you will not be able to connect to the relay server anyway.

You could run this inside a machine image or run the coverage stuff inside the docker container (i.e. without the `-l` option). I would prefer the latter, but the former may be easier to do.

berndbohmeier:
Ah, I thought for some reason we are on a machine image in the clientlib. 
How would I get out the coverage stuff if I run the e2e tests in docker, do you have a hint?

schmir:
> Ah, I thought for some reason we are on a machine image in the clientlib.

The first version from Dong-Ha was using a machine image. But since we started running the end2end tests in the other repos, I didn't want to require it in all other repos as well.

> How would I get out the coverage stuff if I run the e2e tests in docker, do you have a hint?

docker cp. AFAICT you already figured that out.

berndbohmeier:
@dakingha69 This reports only 70%, I was expecting way more and on circleCi it also says for the e2e tests alone already 80%, do you have an idea what is going wrong?

dakingha69:
> @dakingha69 This reports only 70%, I was expecting way more and on circleCi it also says for the e2e tests alone already 80%, do you have an idea what is going wrong?

It seems that `nyc` has additional artefact directories to `.nyc-output` such as
- `<project-root>/node_modules/.cache/nyc`
- `<project-root>/coverage`

See https://github.com/istanbuljs/nyc#setting-the-project-root-directory

Maybe we also need to `docker cp` them? :thinking:

berndbohmeier:
AFAIK it does not create the coverage yet, that is then created by `nyc report` and the caching should not be needed. Maybe that assumptions are wrong.

PR #214, opened 2019-03-30
Example Web3Signer App

### LINKED ISSUE #200: Implement minimal website example for triggering signing with MetaMask
With `Web3Signer` we provide the compatibility with web3 providers such as MetaMask. We should therefore include a small example in `/tests/browser/index.html` to test/showcase this.

dakingha69:
Closes #200 

Also fixes a bug which I discovered while writing the example app and testing the `Web3Signer` properly. As we used `this.user.address` for creating transactions, etc. in multiple places, it threw an error when using an injected web3 provider via MetaMask because `this.user.address` was only set when creating a wallet via `ethers`. Therefore I replaced all references to `this.user.address` with `await this.user.getAddress()`.


cducrest:
I had to install `http-server` to make it work.
Could not get network addresses, but that is potentially because I do not remember how to.

dakingha69:
@cducrest I added a `only_backend` flag to the end2end repository. See https://github.com/trustlines-network/end2end/tree/feature/only-backend-option. If you checkout the branch and then run
```bash
./run-e2e.sh -b
```
you can connect MetaMask to Parity and it should work.

dakingha69:
> What about adding http-server to some dev requirements?

Yes it is already added here: https://github.com/trustlines-network/clientlib/blob/b519c3782ec77742058a29ab9818e5c70ee3ffcf/package.json#L69

PR #226, opened 2019-05-24
Change link base and add optional custom link base

### LINKED ISSUE #224: Switch to using custom uri scheme instead of normal http url
We will switch to the custom scheme `trustlines://` because using the `http://trustlines.network` would require every app to have access to that domain.

dakingha69:
Based on #225 

Closes #224

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -119,9 +119,10 @@ export const buildUrl = (baseUrl: string, params?: any): string => {
       .filter(key => params[key])
       .reduce(
         (acc, paramKey) =>
-          `${acc}/${
-            baseUrl.indexOf('?') === -1 ? '?' : '&'
-          }${paramKey}=${encodeURIComponent(params[paramKey])}`
+          `${acc}${
+            acc.indexOf('?') === -1 ? '?' : '&'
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -108,28 +108,33 @@ export const websocketStream = (
...
- * Returns a trustlines.network link.
- * @param params parameters for link
+ * Returns a `trustlines://` link.
+ * @param params Parameters of link.
+ * @param customBase Optional custom base instead of `trustlines://`.
  */
-export const createLink = (params: any[]): string => {
-  const base = 'http://trustlines.network/v1'
+export const createLink = (params: any[], customBase?: string): string => {
+  const base = customBase || 'trustlines:/'
```
berndbohmeier: Is this because the params part will start with a `/`? I would argue that this path is relative, so it should not start with / and it would also make using this function more clear, because I would expect I can give it `myprefix://` but right now I would need to give `myprefix:/`.

berndbohmeier:
Just change that it can handle `customprefix://` or `http://mydomain.com` or `http://test.com/myapi/` It could just check if the last character is a / and add it if it is missing.

PR #229, opened 2019-06-21
Feature/return sent payment request

### LINKED ISSUE #228: Add return value to messaging.paymentRequest
Currently the sender of a payment request via the interface `messaging.paymentRequest` is not able to somehow save this payment request because the method does not return something. If the payment request was successfully sent, the method should return the payment request.

dakingha69:
Closes #228

### DIFF
--- a/src/Messaging.ts
+++ b/src/Messaging.ts
@@ -22,44 +22,54 @@ export class Messaging {
     this.provider = params.provider
   }
 
+  /**
+   * Sends a payment request to given `counterParty` and returns created payment request.
+   * @param networkAddress Address of currency network.
+   * @param counterPartyAddress Address of counter party.
+   * @param value Requested payment amount.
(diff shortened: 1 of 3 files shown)


src/Messaging.ts
```diff
@@ -22,44 +22,54 @@ export class Messaging {
     this.provider = params.provider
   }
 
+  /**
+   * Sends a payment request to given `counterParty` and returns created payment request.
+   * @param networkAddress Address of currency network.
+   * @param counterPartyAddress Address of counter party.
+   * @param value Requested payment amount.
+   * @param subject Optional subject of payment request.
+   */
```
berndbohmeier: Describe also the return value.

berndbohmeier: So what fields are in the returned payment request.

PR #231, opened 2019-07-04
Add formatting of amount in sent payment request

dakingha69:
Previously only the raw value was sent. This lead to inconsistencies as every numerical value returned from the clientlib should be formatted to `Amount`.

### DIFF
--- a/src/Messaging.ts
+++ b/src/Messaging.ts
@@ -44,7 +44,10 @@ export class Messaging {
       networkAddress,
       from: await this.user.getAddress(),
       to: counterPartyAddress,
-      amount: utils.calcRaw(value, decimals.networkDecimals).toString(),
+      amount: utils.formatToAmount(
+        utils.calcRaw(value, decimals.networkDecimals),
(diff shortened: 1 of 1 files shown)


berndbohmeier:
What are you trying to fix here?
This way now messaging and events are handled differently (messaging amount object, events only raw value), do we really want that?

dakingha69:
> What are you trying to fix here?

The bug with the non-formatted amount in the payment request was due to this.

> This way now messaging and events are handled differently (messaging amount object, events only raw value), do we really want that?

Ah get your point... Didn;t think of events at the moment. But previously the amount sent via a payment request (messaging) was the raw value string. And as I recall, every "value" returned from the clientlib  to the outside should be a formatted amount object? I would maybe argue that events are initally triggered by the relay server and therefore can be in a raw format while with messaging the clientlib itself outputs / creates the message so it has to be formatted.

berndbohmeier:
Lets have a chat about this, I would rather like to keep them the same.

Issue #232, opened 2019-07-24
Add extraData to transfers


cducrest:
In `Paymnet.ts`, we need to add the option for user to specify `extraData` on preparing a TL payment.

We should also add this to `getPath` since it returns the gas estimation for the transaction using this path and this should depend on the (size of) extraData even though it feels odd.
See: https://github.com/trustlines-protocol/relay/blob/d2b44ac488341b4d7ebfaffd4f5d9c905da07b5c/docs/RelayAPI.md#transfer-path-in-currency-network

berndbohmeier:
Maybe we can improve on the naming of `getPath`

PR #234, opened 2019-07-25
Feature/provide extra data on transfers

cducrest:
closes: https://github.com/trustlines-protocol/clientlib/issues/232
Works in pair with: https://github.com/trustlines-protocol/relay/pull/300 and https://github.com/trustlines-protocol/contracts/pull/208

Do not merge this, will base another PR on top of it for event and needs testing.

### DIFF
--- a/src/Payment.ts
+++ b/src/Payment.ts
@@ -59,12 +59,14 @@ export class Payment {
    * @param options.gasPrice Custom gas price.
    * @param options.gasLimit Custom gas limit.
    * @param options.feePayer Either `sender` or `receiver`. Specifies who pays network fees.
+   * @param extraData Extra data that will appear in the Transfer event when successful
    */
   public async prepare(
     networkAddress: string,
     receiverAddress: string,
     value: number | string,
-    options: PaymentOptions = {}
+    options: PaymentOptions = {},
+    extraData: string = '0x'
   ): Promise<PaymentTxObject> {
     const { gasPrice, gasLimit, networkDecimals } = options
     const decimals = await this.currencyNetwork.getDecimals(networkAddress, {
@@ -78,7 +80,8 @@ export class Payment {
       {
         ...options,
         networkDecimals: decimals.networkDecimals
-      }
+      },
+      extraData
     )
     if (path.length > 0) {
       const {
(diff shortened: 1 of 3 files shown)


src/Payment.ts
```diff
@@ -164,7 +168,8 @@ export class Payment {
     senderAddress: string,
     receiverAddress: string,
     value: number | string,
-    options: PaymentOptions = {}
+    options: PaymentOptions = {},
+    extraData = '0x'
```
berndbohmeier: I would add to the documentation that it will return the path and the estimated fees and gas. Then it is maybe not so surprising that it takes the extraData. 
Suggestion: rename it to `getTransferPathInfo` or similar?

src/Payment.ts
```diff
@@ -59,12 +59,14 @@ export class Payment {
...
    * @param options.feePayer Either `sender` or `receiver`. Specifies who pays network fees.
+   * @param extraData Extra data that will appear in the Transfer event when successful
    */
   public async prepare(
     networkAddress: string,
     receiverAddress: string,
     value: number | string,
-    options: PaymentOptions = {}
+    options: PaymentOptions = {},
+    extraData: string = '0x'
```
berndbohmeier: This is better part of `PaymentOptions` then you dont have to do the `undefined` hack you do in the e2e tests. 
There are no named parameters in typescript, so we use the object approach. This should be the last argument.
See https://medium.com/@afontcu/cool-javascript-9-named-arguments-functions-that-get-and-return-objects-337b6f8cfa07 for mor info.

src/Payment.ts
berndbohmeier: same as above.

cducrest: I am fine renaming but I thought renaming on things that are exported in the clientlib is annoying to the mobile app?

berndbohmeier: There is only one file where we access stuff of the clientlib so it is easy to check. https://github.com/trustlines-network/mobileapp/blob/develop/src/shared/lib/client.js
We do not use it directly.

cducrest: Fixed in other PR

src/Payment.ts
cducrest: Fixed in other PR

src/Payment.ts
cducrest: Fixed in other PR

Issue #242, opened 2019-08-26
Add function to freeze trustlines


weilbith:
send a request for freezing the trustline in the network
(other party need to do the same to accept it)

Issue #241, opened 2019-08-26
Add frozen status to trustlines


weilbith:
- shows if trustlines is frozen
- shows if there is a request from the other participant to freeze the trustline

PR #243, opened 2019-08-30
Feature/add frozen trustlines status

cducrest:
Closes: https://github.com/trustlines-protocol/clientlib/issues/241
closes: https://github.com/trustlines-protocol/clientlib/issues/242

Add information as to whether a trustline is frozen. Since it was not too much work, also add ability to freeze a trustline with `prepareUpdate()`

See https://github.com/trustlines-protocol/relay/pull/313 for updated relay api with frozen TL info
See https://github.com/trustlines-protocol/contracts/pull/233 for new CurrencyNetwork TL update function signatures with `isFrozen` on `updateTrustline()`

### DIFF
--- a/src/Trustline.ts
+++ b/src/Trustline.ts
@@ -63,6 +63,7 @@ export class Trustline {
    *                See type `TrustlineUpdateOptions` for more information.
    * @param options.interestRateGiven Proposed interest rate given by initiator to counterparty in % per year.
    * @param options.interestRateReceived Proposed interest rate received by initiator from counterparty in % per year.
+   * @param options.isFrozen Whether we propose to freeze the trustline.
    * @param options.networkDecimals Decimals of currency network can be provided manually if known.
    * @param options.interestRateDecimals Decimals of interest rate in currency network can be provided manually if known.
    * @param options.gasLimit Custom gas limit.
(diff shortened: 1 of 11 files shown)


src/Trustline.ts
```diff
@@ -83,6 +84,7 @@ export class Trustline {
       gasLimit,
       gasPrice
     } = options
+    const isFrozen = options.isFrozen
```
schmir-at-bb: Shouldn't this be moved into the `const {...}` block above?

src/Trustline.ts
```diff
@@ -108,21 +110,39 @@ export class Trustline {
...
+      interestRateGiven !== undefined ||
+      interestRateReceived !== undefined ||
+      isFrozen !== undefined
+    ) {
+      if (
+        interestRateGiven !== undefined &&
+        interestRateReceived !== undefined &&
+        isFrozen !== undefined
+      ) {
+        updateFuncName = 'updateTrustline'
```
schmir-at-bb: Please make this if statement 'flat':
```
If (    interestRateGiven !== undefined &&
        interestRateReceived !== undefined &&
        isFrozen !== undefined) {
...} else if (      interestRateGiven !== undefined ||
      interestRateReceived !== undefined ||
      isFrozen !== undefined
    ) {
throw ...
}
```

src/Trustline.ts
cducrest: yes

src/Trustline.ts
cducrest: Ok

PR #247, opened 2019-09-23
Add default value of false for isFrozen

cducrest:
Add default value of false for isFrozen for prepareUpdate (and prepareAccept) when interests are set as a quick fix to be removed later

to allow proper functionning of mobile apps built on top of it.
Needs to be reconsidered after discussion on final solution.


compojoom:
Do we really need to do this in the clientlib? If isFrozen is required, then it is required. I udpated the app to provide the isFrozen flag (it is used in 2 places)

PR #248, opened 2019-09-24
resolve fsevents to 1.2.9 to fix a build issue

compojoom:
When trying to run yarn on node 12 we get an error:
```
make: *** [Release/obj.target/fse/fsevents.o] Error 1
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
```

We don’t directly use fsevent, but chokidar does. Watchpack includes chockider and webpack includes watchpack…

Watchpack in version 2 (still in beta) no longer uses chokidar, but we are stuck with 1.6 - this is what webpack has specified. So to fix the issue with building on node v12 I’ve provided a resolve param in the package.json that specifies that we want to use fsevent 1.2.9 and that version builds properly on node v12

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,9 +86,6 @@
     "webpack": "^4.16.0",
     "webpack-cli": "^3.0.8"
   },
-  "resolutions": {
-    "fsevents": "1.2.9"
-  },
   "dependencies": {
     "bignumber.js": "7.2.1",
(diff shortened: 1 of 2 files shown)


berndbohmeier:
This fails with the message
`error fsevents@1.2.9: The platform "linux" is incompatible with this module.
error Found incompatible module.`
Can you check this?

PR #254, opened 2019-10-04
Feature/deterministic identity address

cducrest:
I could successfully run the e2e tests locally with:
relay version: https://github.com/trustlines-protocol/relay/pull/329
and contracts: https://github.com/trustlines-protocol/contracts/pull/254

I am not sure how to handle the addresses.json file in `tests/e2e-config`. I wish for it to be there so that unit tests can easily be ran with sample values but it does not have to be committed since it is erased by local e2e tests. Also I can not make a comment to the file to explain where it comes from.

The recovery of identity wallet from seed does not handle the case where the identity is not yet deployed. This case is also not handled in the e2e tests. We need to fix this eventually.

I could not successfully deploy identities in parallel: `await Promise.all(tl1.user.deployIdentity(), tl2.user.deployIdentity())` would throw an error like `Delegate did not deploy the right identity contract. Deployed '0x1234...' instead of '0xabcd...'`. I assume this could be due to the fact that the response from the relay for the deployment of one identity goes to the wrong user that expects a different address. I am not sure whether this needs to be handled and how.


berndbohmeier:
> I am not sure how to handle the addresses.json file in tests/e2e-config. I wish for it to be there so that unit tests can easily be ran with sample values but it does not have to be committed since it is erased by local e2e tests. Also I can not make a comment to the file to explain where it comes from.

We could put a readme in the folder?

src/wallets/EthersWallet.ts
```diff
@@ -76,6 +76,18 @@ export class EthersWallet implements TLWallet {
...
 
+  /**
+   * Deploys a new identity contract on the chain
+   */
+  public async deployIdentity(): Promise<string> {
+    // This does not have to deploy any identity, because it does not have it
+    return this.address
+  }
+
+  public async isIdentityDeployed(): Promise<boolean> {
```
berndbohmeier: should we not rather return true? 
in a app flow it would probably do something like 
```
if not isIdentityDeploy
   deployIdentity()
```

so if it uses no identity wallets it should maybe just return true, as in all is good?

cducrest: Seems abusive, but sure.

src/wallets/IdentityWallet.ts
```diff
@@ -61,33 +73,71 @@ export class IdentityWallet implements TLWallet {
...
+  public async isIdentityDeployed(): Promise<boolean> {
+    const response = await this.provider.fetchEndpoint<any>(
+      `/identities/${this.address}`
     )
-
-    return {
-      address: identity.identity,
-      pubKey: 'Not implemented yet',
-      serializedWallet
+    if ('not found' in response) {
```
berndbohmeier: This will not work as this.provider.fetchEndpoint will already throw an error. 
You will have to use try catch

src/wallets/IdentityWallet.ts
```diff
@@ -116,7 +166,7 @@ export class IdentityWallet implements TLWallet {
       typeof progressCallback === 'function' && progressCallback
     )
 
-    this.identityAddress = identityAddress
+    this.identityAddress = toChecksumAddress(identityAddress)
```
berndbohmeier: that is a weird position to do that?
we should already save it as checksummed

berndbohmeier: probably we should fix calculateIdentityAddress

berndbohmeier: Is already fixed, so just remove it here.

berndbohmeier: No, lets leave it.

PR #258, opened 2019-10-23
Feature/non mandatory keystore encryption

### LINKED ISSUE #257: Make use of encrypted ethersKeystore in account creation / load optional
## Problem
We currently make the use of the encrypted ethers keystore mandatory for creating and loading accounts in the clientlib. This limits the use of the library. In the mobileapp, for example, we want to skip the keystore encryption process and also use the mnemonic to load an account which is in this combination not possible right now.

## Solution
- Remove encryption and serialization of ethers keystore for creating and loading an account and move into separate method, e.g. here https://github.com/trustlines-protocol/clientlib/blob/master/src/wallets/IdentityWallet.ts#L81-L89
- Rethink our own wallet schema types https://github.com/trustlines-protocol/clientlib/blob/master/src/typings.ts#L423-L434
- Remove `serializedWallet` from `UserObject` type https://github.com/trustlines-protocol/clientlib/blob/master/src/typings.ts#L405-L409
- Refactor `loadAccount` such that it can handle different account recovery methods and is not tied to `ethers.Wallet.fromEncryptedJson` https://github.com/trustlines-protocol/clientlib/blob/master/src/wallets/EthersWallet.ts#L144

dakingha69:
Closes #257 

- `wallet.createAccount()` do not load and encrypt the created wallet, instead only returns the created `TLWallet`
- `wallet.loadAccount()` do not require a serialized encrypted wallet as a parameter, instead expect a `TLWallet` -> no wallet decryption required
- `wallet.recoverFromSeed()` / `wallet.recoverFromPrivateKey()` do not require a serialized encrypted wallet as a parameter, instead only the mnemonic / private key -> no wallet decryption required
- `wallet.recoverFromEncryptedWallet()` as additional recovery method
- `wallet.encryptWallet()` as additional method to encrypt wallet with password


src/User.ts
```diff
@@ -125,6 +102,25 @@ export class User {
...
+   */
+  public async encryptWallet(
+    tlWallet: TLWalletSchema,
+    password?: string | ((progress: number) => void),
+    progressCallback?: (progress: number) => void
+  ): Promise<string> {
+    return this.wallet.encryptWallet(
+      tlWallet,
+      typeof password === 'string' ? password : this.defaultPassword,
+      typeof password === 'function' ? password : progressCallback
```
compojoom: Are you sure about all this?

typeof password === 'string' => typeof '' === 'string' - do we want to use empty string as password?

the next line typeof password === 'function' did you mean progressCallback?

src/typings.ts
```diff
@@ -420,17 +420,46 @@ export interface Signature {
...
+export interface EncryptedTLWalletSchema extends TLWalletSchema {
+  meta: {
+    ethersKeystore: string
+  }
+}
+
+export interface EncryptedIdentityWalletSchema extends TLWalletSchema {
+  meta: {
+    ethersKeystore: string
+    identityAddress: string
```
berndbohmeier: I would remove the identityAddress here, we can just recalculate the address here, because it is now deterministic.

src/wallets/EthersWallet.ts
```diff
@@ -129,10 +140,12 @@ export class EthersWallet implements TLWallet {
     }
 
     if (!('TLWalletVersion' in deserializedWallet)) {
-      // Use the old serialising method if no TL_WALLET_VERSION key in the deserialized wallet
-      encryptedKeystore = serializedWallet
+      // Use the old serializing method if no TL_WALLET_VERSION key in the deserialized wallet
```
berndbohmeier: If you want to be able to load the old format for backwards compatibility, it should be in loadAccount, otherwise I would remove it.

src/wallets/EthersWallet.ts
```diff
@@ -89,34 +72,62 @@ export class EthersWallet implements TLWallet {
...
+  }
 
-    const serializedWallet: string = JSON.stringify(deserializedWallet)
+  /**
+   * Encrypts and serializes the given wallet.
+   * @param ethersWallet `TLWallet` of type `WALLET_TYPE_ETHERS`.
+   * @param password Password to encrypt wallet with.
+   * @param progressCallback Optional encryption progress callback.
+   */
+  public async encryptWallet(
```
berndbohmeier: I understood it that we made this function to return the standard format, so we should just return the encrypted keystore here.

src/typings.ts
```diff
@@ -420,17 +420,46 @@ export interface Signature {
...
+
+export interface TLWalletSchema {
   TLWalletVersion: number
-  ethersKeystore: string
+  walletType: WalletType
+  meta?: any
+}
+
+interface EthersWalletMeta {
+  walletFromEthers: ethers.Wallet
```
berndbohmeier: I thought we aggreed on saving the SigningKey, which consists of the mnemonic, the derivation path and the private key, now?

berndbohmeier: I would not like to expose a type which is not from us, this breaks abstraction.

src/User.ts
dakingha69: > typeof password === 'string' => typeof '' === 'string' - do we want to use empty string as password?

It is possible with ethers js, so I would also allow it.

> the next line typeof password === 'function' did you mean progressCallback?

Wanted to give user the option to specify a callback without a password like
```js
user.encryptWallet(tlWallet, progressCallback => {...})
```
But not sure about this...

src/wallets/EthersWallet.ts
```diff
@@ -53,26 +49,13 @@ export class EthersWallet implements TLWallet {
...
-
-    const serializedWallet = this.serializeWallet(encryptedKeystore)
-
+  public async createAccount(): Promise<UserObject<EthersWalletSchema>> {
+    const walletFromEthers = ethers.Wallet.createRandom()
     return {
-      address: this.address,
-      pubKey: this.pubKey,
-      serializedWallet
+      address: await walletFromEthers.getAddress(),
```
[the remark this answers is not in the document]
dakingha69: ok will change this

src/wallets/EthersWallet.ts
dakingha69: Ok so it would be okay for you if I omit backwards compatibility for now? So I would keep version 1 of the TLWallet but include breaking changes as we have no real user currently anyways.

src/typings.ts
dakingha69: Misunderstood something. Will changes this

src/wallets/EthersWallet.ts
dakingha69: changed it so that it only returns the string as returned by `ethers.Wallet.encrypt`

src/wallets/EthersWallet.ts
```diff
@@ -89,34 +72,62 @@ export class EthersWallet implements TLWallet {
...
+    return this.serializeWallet(encryptedKeystore)
+  }
 
-    return serializedWallet
+  /**
+   * Loads given ethers wallet.
+   * @param ethersWallet `TLWallet` of type `WALLET_TYPE_ETHERS`.
+   */
+  public async loadAccount(ethersWallet: EthersWalletSchema): Promise<void> {
+    this.walletFromEthers = ethersWallet.meta.walletFromEthers
```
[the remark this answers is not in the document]
dakingha69: now added to `loadAccount`. it was in `recoverFromEncryptedJson` before

src/typings.ts
```diff
@@ -420,17 +420,35 @@ export interface Signature {
...
+  meta: EthersWalletMeta
 }
 
-export interface EthersWalletSchema {
-  TLWalletVersion: number
-  ethersKeystore: string
-  walletType: string
+export interface IdentityWalletSchema extends TLWalletSchema {
+  type: WalletTypeIdentity
+  meta: EthersWalletMeta
```
berndbohmeier: I would here also store the address, so that you dont have to recalculate the address.

src/typings.ts
```diff
@@ -420,17 +420,35 @@ export interface Signature {
...
+export interface TLWalletSchema {
+  version: number
+  type: WalletType
+  meta?: any
+}
+
+export interface SigningKey {
+  privateKey: string
+  mnemonic: string
+  derivationPath: string
```
berndbohmeier: Thinking about it, we dont have to store the derivation path, because we currently also do not support using in in recover from mnemonic. So we can just hard code it as we do it in recover from mnemonic.

src/utils.ts
```diff
@@ -389,6 +393,33 @@ export const trimUrl = (url: string): string => {
...
+/**
+ * Takes a `TLWallet` of type `ethers` or `identity` and returns an instance of `ethers.Wallet`.
+ * @param wallet `TLWallet` of type `ethers` or `identity`.
+ */
+export function getWalletFromEthers(
+  wallet: EthersWalletSchema | IdentityWalletSchema
+): ethers.Wallet {
+  const { signingKey } = wallet.meta
+  return signingKey.mnemonic && signingKey.derivationPath
+    ? ethers.Wallet.fromMnemonic(signingKey.mnemonic, signingKey.derivationPath)
```
berndbohmeier: Isnt that too slow? isnt the private key in that case recalculated from the mnemonic? We should always recover from the private key directly and additionally restore the mnemonic.

src/wallets/EthersWallet.ts
```diff
@@ -53,27 +54,11 @@ export class EthersWallet implements TLWallet {
...
-
-    const serializedWallet = this.serializeWallet(encryptedKeystore)
-
-    return {
-      address: this.address,
-      pubKey: this.pubKey,
-      serializedWallet
-    }
+  public async createAccount(): Promise<UserObject<EthersWalletSchema>> {
+    const walletFromEthers = ethers.Wallet.createRandom()
```
berndbohmeier: Where do you store the wallet?

src/wallets/EthersWallet.ts
```diff
@@ -89,121 +74,77 @@ export class EthersWallet implements TLWallet {
...
-
-    return {
-      address: this.address,
-      pubKey: this.pubKey,
-      serializedWallet
-    }
+    privateKey: string
+  ): Promise<UserObject<EthersWalletSchema>> {
+    const walletFromEthers = new ethers.Wallet(privateKey)
+    return this.getAccountWithEthersWallet(walletFromEthers)
```
berndbohmeier: where do you store the wallet?

src/wallets/IdentityWallet.ts
```diff
@@ -150,128 +128,86 @@ export class IdentityWallet implements TLWallet {
...
-    const encryptedKeystore = deserializedWallet.ethersKeystore
-    const identityAddress = deserializedWallet.identityAddress
-
-    this.wallet = await ethers.Wallet.fromEncryptedJson(
-      encryptedKeystore,
-      password,
-      typeof progressCallback === 'function' && progressCallback
+    const walletFromEthers = utils.getWalletFromEthers(identityWallet)
+    this.walletFromEthers = walletFromEthers
+    this.identityAddress = calculateIdentityAddress(
```
berndbohmeier: I would not recalculate the address here, because that might be slow. 
Lets just store it.

tests/Fixtures.ts
```diff
@@ -39,27 +42,16 @@ export const parametrizedTLNetworkConfig = [
...
-
 export const extraData = '0x12ab34ef'
 
-export function createUsers(tlInstances: TLNetwork[]) {
-  return Promise.all(tlInstances.map(tl => tl.user.create()))
+export async function createAndLoadUsers(tlInstances: TLNetwork[]) {
+  return Promise.all(
+    tlInstances.map(async tl => {
+      const createdUser = await tl.user.create()
+      await tl.user.load(createdUser.wallet)
```
berndbohmeier: So you split up the creation and loading of an account on purpose? Why? I think it is intuitive that create account, restore account will not change the state, but load will.

dakingha69:
I changed the naming into `credentials`. So we have now smth like `wallet.createCredentials`, `wallet.recoverCredentialsFrom*`, `wallet.loadCredentials`, etc. I also removed the type `UserObject` as this was unnecessary and misleading. Therefore I added the property `address` directly to the type `TLWalletCredentials`.

tests/unit/User.test.ts
```diff
@@ -73,16 +71,13 @@ describe('unit', () => {
       beforeEach(() => init())
 
       it('should load a user from serialized wallet', async () => {
-        const loadedUser = await user.load(keystore1)
-        assert.hasAllKeys(loadedUser, ['address', 'serializedWallet', 'pubKey'])
-        assert.isString(loadedUser.address)
-        assert.isString(loadedUser.serializedWallet)
-        assert.isString(loadedUser.pubKey)
+        const loadedUser = await user.loadFrom(TL_WALLET_DATA)
```
berndbohmeier: this should rather test that it is loaded into the library, so something like checking the address would be good.

berndbohmeier:
Can be merged after you addressed the small problems if you need it, but there are still performance problems. 
But we should try to fix this one last issue, that we do everything twice. 
So we generate the ethers wallet and then take the data from it to generate it again. This is slow.

src/wallets/WalletFromEthers.ts
```diff
@@ -0,0 +1,84 @@
...
+    walletData: EthersWalletData | IdentityWalletData
+  ) {
+    const { signingKey } = walletData.meta
+    const { privateKey, mnemonic } = signingKey
+    return new this(privateKey, mnemonic)
+  }
+
+  public static createRandom() {
+    const { privateKey, mnemonic } = super.createRandom()
+    return new this(privateKey, mnemonic)
```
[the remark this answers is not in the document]
berndbohmeier: But I checked, instantiation should be fast, so probably does not matter.

package.json
```diff
@@ -13,8 +13,8 @@
     "prettier:big-tabs": "prettier --write --print-width 120 --no-semi --single-quote --tab-width 4  './{src,tests}/**/*.ts'",
     "test": "mocha --timeout 30000 --require isomorphic-fetch -r ts-node/register tests/**/**.test.ts",
     "test:e2e": "nyc mocha --timeout 30000 --require isomorphic-fetch -r ts-node/register tests/e2e/**.test.ts",
-    "test:integration": "nyc mocha --timeout 5000 --require isomorphic-fetch -r ts-node/register tests/integration/**.test.ts",
-    "test:unit": "nyc mocha --timeout 5000 --require isomorphic-fetch -r ts-node/register tests/unit/**.test.ts",
+    "test:integration": "nyc mocha --timeout 30000 --require isomorphic-fetch -r ts-node/register tests/integration/**.test.ts",
```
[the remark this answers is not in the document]
dakingha69: ups no only unit tests

PR #264, opened 2019-11-13
Add method to retrieve version of configured relay server

dakingha69:
Adds `getRelayVersion` to `RelayProvider` so that user can fetch the version of current configured relay server.

### DIFF
--- a/src/providers/RelayProvider.ts
+++ b/src/providers/RelayProvider.ts
@@ -102,6 +102,14 @@ export class RelayProvider implements TLProvider {
   }
 
   /**
+   * Returns the version of the currently configured relay server.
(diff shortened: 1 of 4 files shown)


src/providers/RelayProvider.ts
```diff
@@ -101,6 +101,14 @@ export class RelayProvider implements TLProvider {
     return utils.formatToAmount(utils.calcRaw(balance, 18), 18)
   }
 
+  /**
+   * Returns the version of the currently configured relay server.
+   * @returns Version of relay in the format `relay/vX.X.X`.
```
berndbohmeier: the idea was that it is `<name>/vX.X.X`, so for a different implementation is might say `myAwsomeRustRelay/v0.2.2`

Issue #267, opened 2019-11-18
Ask for fee / use the given fee for meta-tx


cducrest:
User should ask what the fee should be to the relay server when preparing a transaction, and include that fee when signing it.

Issue #253, opened 2019-10-02
Add delegation fees and network of fees as tx options for meta transactions


dakingha69:
When executing a normal transaction we are able to pass in optional `gasPrice` and `gasLimit` parameters in `TxOptions` when preparing it, e.g. https://github.com/trustlines-protocol/clientlib/blob/master/src/Transaction.ts#L53-L54. Those values are then used to return the transaction costs in `ethFees`.
Currently we do the same for meta transactions which is misleading because in this context the parameters `gasPrice` and `gasLimit` are not needed. Instead we need `delegationFees` and `networkOfFees` as  additional optional parameters. If not set by the user those should also default to values returned by the relay server. Instead of `ethFees` the `delegationFees` and `networkOfFees` should also be returned.

cducrest:
done.

PR #269, opened 2019-11-20
Feature/ask meta tx fees

cducrest:
I decided to ask fro delegation fees when preparing a contract transaction instead of creating an extra step in between preparation and confirmation of transaction.

In `prepareContractTransaction` I could check what signer is used and only get meta transaction fees if the signer is an IdentityWallet (https://github.com/trustlines-protocol/clientlib/pull/269/files#diff-1e657e773ebb13137b9281cc47cdd8efR65). This would allow to remove `getMetaTxFees()` from `EthersWallet` but I was not bothered by it since there are already methods related to identity in there (e.g. `deployIdentity()`).

Closes https://github.com/trustlines-protocol/clientlib/issues/267
the e2e tests will not pass unless this is merged: https://github.com/trustlines-protocol/relay/pull/356

### DIFF
--- a/src/Transaction.ts
+++ b/src/Transaction.ts
@@ -13,7 +13,8 @@ import {
   TxObjectInternal,
   TxOptionsInternal
 } from './typings'
-import { IdentityWallet } from './wallets/IdentityWallet'
+
+import { CurrencyNetwork } from './CurrencyNetwork'
 
 const ETH_DECIMALS = 18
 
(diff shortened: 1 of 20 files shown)


cducrest:
Also closes https://github.com/trustlines-protocol/clientlib/issues/253

src/Transaction.ts
```diff
@@ -56,7 +65,18 @@ export class Transaction {
...
+      const metaTransactionFees: MetaTransactionFees = await this.signer.getMetaTxFees(
+        rawTx
+      )
+      rawTx.delegationFees = metaTransactionFees.delegationFees
+      rawTx.currencyNetworkOfFees = metaTransactionFees.currencyNetworkOfFees
+    }
+
+    const ethFees = new BigNumber(rawTx.gasLimit).multipliedBy(rawTx.gasPrice)
     return {
       ethFees: utils.formatToAmountInternal(ethFees, ETH_DECIMALS),
```
berndbohmeier: Here we return the eth fees, so we should here probably also return the delegation fees. 
The same for the prepare methods in transfer and trustline.

cducrest:
Should be fine now.
The only places where there are no delegation fees are within `Exchange.ts` (e.g. `prepTakeOrder`) and `EthWrapper.ts` (e.g. `prepTransfer`). Should I add it there too?

PR #276, opened 2020-01-03
Remove install of yarn when building docker image

cducrest:
It is already installed and will raise an error now https://
github.com/nodejs/docker-node/commit/
d46f9eb95a2db1
2ea29388b3c9b39ba81efe0f51#diff-04c6e90faac2675aa89e2176d2eec7d8

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,6 @@
 #   docker build . -t e2e
 
 FROM node:10-jessie as builder
-RUN npm install -g yarn
 
 COPY . /clientlib
 WORKDIR /clientlib


cducrest:
> Strictly speaking newlines inside URLs aren't allowed. It makes little sense to have these URLs in the git log if you can't click them. Please fix your editor setup.

So should I have lines bigger than whatever the limit we set for the commit message?

schmir-at-bb:
> So should I have lines bigger than whatever the limit we set for the commit message?

yes, it's the sensible thing to do.

Issue #280, opened 2020-01-06
UnhandledPromiseRejectionWarning: AssertionError but end2end tests do succeed


schmir-at-bb:
I see a lot of `UnhandledPromiseRejectionWarning: AssertionError: expected { Object (ethFees, delegationFees, ...) } to have keys 'rawTx', and 'ethFees'` errors when running the end2end tests, but they do not fail and instead report success.

```
e2e             | 2020-01-06T12:40:26.550666413Z     at getBasePromise.then.then.newArgs (/clientlib/node_modules/chai-as-promised/lib/chai-as-promised.js:302:22)
e2e             | 2020-01-06T12:40:26.550688247Z     at process._tickCallback (internal/process/next_tick.js:68:7)
e2e             | 2020-01-06T12:40:26.550694291Z (node:42) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 6)
e2e             | 2020-01-06T12:40:26.554704024Z         ✓ should return txId for network WITHOUT interest rates (165ms)
e2e             | 2020-01-06T12:40:28.598318800Z         ✓ should return txId for network with DEFAULT interest rates (38ms)
e2e             | 2020-01-06T12:40:30.665683021Z         ✓ should return txId for network with CUSTOM interest rates (61ms)
e2e             | 2020-01-06T12:40:32.719588422Z         ✓ should return txId for trustline update request with freezing (48ms)
e2e             | 2020-01-06T12:40:34.724658433Z       #getRequests()
e2e             | 2020-01-06T12:40:36.804404329Z         ✓ should return latest request for network WITHOUT interest rates (2080ms)
e2e             | 2020-01-06T12:40:38.881740064Z         ✓ should return latest request for network with DEFAULT interest rates (2076ms)
e2e             | 2020-01-06T12:40:40.971100250Z         ✓ should return latest request for network with CUSTOM interest rates (2089ms)
e2e             | 2020-01-06T12:40:43.102362581Z         ✓ should return latest request for trustline update request with freezing (2131ms)
e2e             | 2020-01-06T12:40:43.102652659Z       #prepareAccept()
e2e             | 2020-01-06T12:40:43.103159746Z         ✓ should prepare accept tx for network WITHOUT interest rates
[...]

PR #281, opened 2020-01-06
Handle the result of promises in tests

schmir-at-bb:
This fixes test functions using promises, but not waiting for their
result.

We did run into multiple 'UnhandledPromiseRejectionWarning:
AssertionError' errors. Though the test runner did log them, it
ignored them while reporting the result of the tests.

This is bad, since it will hide actual errors in the tests.

Instead of something like:

,----
| it('should return all user addresses of specific currency network', () => {
|   expect(
|     currencyNetwork.getUsers(networks[0].address)
|   ).to.eventually.be.an('array')
| })
`----

we must make the inner function async and await the expect value:

,----
| it('should return all user addresses of specific currency network', async () => {
|   await expect(
|     currencyNetwork.getUsers(networks[0].address)
|   ).to.eventually.be.an('array')
| })
`----

This fixes all occurences that use's chai's `eventually` [1].

See https://github.com/trustlines-protocol/clientlib/issues/280

[1] https://www.chaijs.com/plugins/chai-as-promised/

### DIFF
--- a/tests/e2e/CurrencyNetwork.test.ts
+++ b/tests/e2e/CurrencyNetwork.test.ts
@@ -87,8 +87,8 @@ describe('e2e', () => {
       })
 
       describe('#getUsers()', () => {
-        it('should return all user addresses of specific currency network', () => {
-          expect(
+        it('should return all user addresses of specific currency network', async () => {
+          await expect(
             currencyNetwork.getUsers(networks[0].address)
           ).to.eventually.be.an('array')
         })
--- a/tests/e2e/EthWrapper.test.ts
+++ b/tests/e2e/EthWrapper.test.ts
@@ -41,14 +41,16 @@ describe('e2e', () => {
       })
 
       describe('#getAddresses()', () => {
-        it('should return array of wrapper addresses', () => {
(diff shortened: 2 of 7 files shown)


schmir-at-bb:
@compojoom Can you please have a look at that one? It fixes the problem we talked about yesterday.

compojoom:
I was thinking that we were using jest and I was referring to this yesterday: 
https://jestjs.io/docs/en/asynchronous

But now that I've looked into package.json - we are using mocha with chai. And we've added the chai assertions for promises package:
https://www.chaijs.com/plugins/chai-as-promised/

The async/await syntax makes chai-as-promised obsolete. Maybe we should check if we have tests that are using it and modify them for async/await syntax.

However there is one thing that I dislike in the way you've written the tests.

You have one liners like this:
```
          await expect(
            currencyNetwork.getUsers(networks[0].address)
          ).to.eventually.be.an('array')
```

And they are perfectly fine, but don't you think that it would make reading the tests easier if they were
```
const address = await currencyNetwork.getUsers(networks[0].address)
expect(address).to.eventually.be.an('array')
```
???

PR #282, opened 2020-01-07
Add .dockerignore file

schmir-at-bb:
The build context could get rather large before. Let's remove stuff we
know we don't need.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+node_modules
+.nyc_output
+.git
+tests/e2e-config/addresses.json


cducrest:
Can we also ignore the file generated when running the e2e tests `tests/e2e-config/addresses.json`? I did not achieve to do this.

schmir-at-bb:
> Can we also ignore the file generated when running the e2e tests tests/e2e-config/addresses.json?

yes, but the file is already available in git. I wonder if we need there. You seem to have added it in 348a82c3500718e9117cc ?

PR #296, opened 2020-01-22
Enable ws stream subscribers to handle ws errors

dohaki:
Related to https://github.com/trustlines-network/mobileapp/issues/880.
We basically want to be able to show in the app that it is not connected to a relay server.

It was not possible to handle ws errors when using the clientlib.
Even though the clientlib handles the reconnection via
[reconnecting-websocket](https://github.com/pladaria/reconnecting-websocket),
one could not know or handle that the ws connection got closed.

It is now possible to pass in options for the reconnection or even
disable it via `reconnectOnError: false`.
It is also possible to detect that a ws connection errored while the
clientlib handles the reconnection.

### DIFF
--- a/src/Event.ts
+++ b/src/Event.ts
@@ -14,7 +14,8 @@ import {
   AnyNetworkEventRaw,
   AnyTokenEventRaw,
   DecimalsOptions,
-  EventFilterOptions
+  EventFilterOptions,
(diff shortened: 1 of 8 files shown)


cducrest:
LGTM

ghost (approved):

Issue #293, opened 2020-01-20
Update execute meta-tx signature


cducrest:
Add new fields for gasLimit, gasPrice, baseFee, timeLimit

PR #297, opened 2020-01-24
Update meta-tx to use gasPrice, gasLimit, feeRecipient, and other

cducrest:
https://github.com/trustlines-protocol/clientlib/issues/293


src/Transaction.ts
```diff
@@ -141,37 +161,30 @@ export class Transaction {
...
+      rawTx
+    )
 
-    let decimals = 0
-    if (delegationFees !== '0' && currencyNetworkOfFees !== '') {
-      decimals = (await this.currencyNetwork.getDecimals(currencyNetworkOfFees))
-        .networkDecimals
+    let currencyDecimals = 0
+    if (
+      metaTransactionFees.baseFee !== '0' &&
```
berndbohmeier: We should only check for currencyNetwork as the gasPrice could be non zero.

PR #304, opened 2020-02-10
Feature/md typedoc

dakingha69:
This PR includes the API reference for the clientlib which is auto-generated using [typedoc](https://typedoc.org/) from the comments within the source code. The plugin https://github.com/tom-grey/typedoc-plugin-markdown is used to generate the documentation as md files.

As we only want to expose the most important APIs to the user, I tried to find a way to "whitelist" specific methods/classes with typedoc but failed to do so without having to refactor the whole library. So, for now, I added a table of contents file in `./docs/api-reference-index.md` where we can refer to the classes that should be part of the documentation. All auto-generated files from typedoc are saved under `./docs/api`. If you guys have another suggestion on how to make this better, please let me know.


docs/api-reference-index.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Table of Content
+
+- [`TLNetwork`](./api/classes/_tlnetwork_.tlnetwork.md)
+  - [`CurrencyNetwork`](./api/classes/_currencynetwork_.currencynetwork.md)
+  - [`User`](./api/classes/_user_.user.md)
+  - [`Trustline`](./api/classes/_trustline_.trustline.md)
+  - [`Payment`](./api/classes/_payment_.payment.md)
+  - [`Events`](./api/classes/_events_.events.md)
+- [`Utils`](<(./api/modules/_utils_.md)>)
```
weilbith: That link is broken. Not sure if this is a typo or what was the origin intention by the `<`'s.

docs/api/README.md
```diff
@@ -0,0 +1,27 @@
+# trustlines-clientlib
+
+## Index
+
+### External modules
+
+- ["Contact"](modules/_contact_.md)
```
weilbith: So that's auto-generated. Can we configure it to don't use such ugly quotes? :thinking:

src/CurrencyNetwork.ts
```diff
@@ -31,6 +40,7 @@ export class CurrencyNetwork {
 
   private provider: TLProvider
 
+  /** @hidden */
```
weilbith: Just for me to get this: when do you hide something manually that should not be private in the code? :thinking:

dakingha69: So currently some methods need to be publicly available for internal use within the library source code, such as the constructors for the subclasses, but should not directly be used externally by the user... Can be improved in the future when we refactor the clientlib

docs/api/README.md
dakingha69: After a little bit of digging, I could not find a straightforward way to remove them unfortunately.. Have a look at https://github.com/TypeStrong/typedoc/issues/68

weilbith: Okay then we have to live with that. :man_shrugging: 
Thanks for the additional research!

weilbith:
What is the state of this PR? Since it is the last one that blocks the big story of trustlines-network/project#798

schmir-at-bb:
Do we really want to add the generated markdown files to git?

weilbith:
> Do we really want to add the generated markdown files to git?

I also don't like too much to have sources and generated products in a repo. The problem is that we need that for the centralized dev-docu repo which will fetch the `./doc` folder from all relevant projects. Not sure how to solve that better. Maybe we could have a dedicated branch for that (so it does at least no pollute the "main" branches) and let CI generate that. Then could Tatu still refer this repo as sub-module with a specific branch filter.

dakingha69:
Yes definitely is a lot of clutter the auto-generated md files... I like the dedicated branch approach @weilbith mentioned. So shall I clean up the PR by removing the respective files and only leaving the not auto-generated files and changes in?

Issue #309, opened 2020-02-12
Allow for separate messaging service provider


berndbohmeier:
Currently we mix the core protocol features provided by relay server and the currently app only features (like payment requests, usernames). 
Because the server running the features will now split, we need to different endpoints be called. 
To do that it is enough to add a new provider just for the functions in `Messaging.ts`

PR #310, opened 2020-02-13
Split providers into messaging and trustlines

cducrest:
I put the function `getVersion()` in the base `Provider.ts` even though it assumes the provider implements an endpoint `version` but I felt like it was not so much to assume and would prove useful.
closes: https://github.com/trustlines-protocol/clientlib/issues/309

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -58,10 +58,18 @@ The user therefore needs some coins ([TLC](https://explore.tlbc.trustlines.found
 
 ```javascript
 const laika = new TLNetwork({
-  protocol: 'https',
-  wsProtocol: 'wss',
-  host: 'relay0.testnet.trustlines.network',
-  path: '/api/v1',
+  relayProviderUrlObject: {
+    protocol: 'https',
+    wsProtocol: 'wss',
+    host: 'relay0.testnet.trustlines.network',
+    path: '/api/v1'
+  },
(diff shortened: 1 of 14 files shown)


cducrest:
Should be good now, I made `getApiUrl` and `getWsUrl` functions in `TLNetwork` which we could have put in `utils` but I would argue they are specific to the way we take the input url in TLNetwork (via full url / object with defaults).

docs/guides.md
```diff
@@ -58,10 +58,30 @@ The user therefore needs some coins ([TLC](https://explore.tlbc.trustlines.found
 
 ```javascript
 const laika = new TLNetwork({
-  protocol: 'https',
-  wsProtocol: 'wss',
-  host: 'relay0.testnet.trustlines.network',
-  path: '/api/v1',
+  relayProviderApiUrl: {
+    protocol: 'https',
+    wsProtocol: '',
```
berndbohmeier: Why do we now have wsProtocol in here and also relayProviderWsUrl? that does not make sense. Also the names are super long. An external does not now about provider or so. He probably does not even care that we use websockets or an api at all.
Please just use: `relayApiUrl` and `messagingApiUrl` or even just `relayUrl`.
and if someone provides the url with `https` then wie use `wss` for the websocket, and if he uses `http` we use `ws`

PR #312, opened 2020-02-19
Add identitywallet function for getting tx status

cducrest:
requires: https://github.com/trustlines-protocol/relay/pull/426

### DIFF
--- a/src/providers/RelayProvider.ts
+++ b/src/providers/RelayProvider.ts
@@ -9,6 +9,7 @@ import {
   Amount,
   MetaTransaction,
   MetaTransactionFees,
+  MetaTransactionStatus,
   TxInfos,
   TxInfosRaw
 } from '../typings'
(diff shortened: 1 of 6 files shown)


src/providers/RelayProvider.ts
```diff
@@ -133,6 +137,16 @@ export class RelayProvider implements TLProvider {
     }
   }
 
+  public async getMetaTxStatus(
+    identityAddress: string,
+    metaTransactionHash: string
+  ) {
+    const { MetaTransactionStatus } = await this.fetchEndpoint<any>(
+      `/identities/${identityAddress}/meta-transactions/${metaTransactionHash}`
```
berndbohmeier: here you used `meta-transactions` but did you not use `meta-transaction-status` in the relay. My suggestion would be to use the link here.

src/providers/RelayProvider.ts
```diff
@@ -133,6 +137,16 @@ export class RelayProvider implements TLProvider {
     }
   }
 
+  public async getMetaTxStatus(
+    identityAddress: string,
+    metaTransactionHash: string
+  ) {
+    const { MetaTransactionStatus } = await this.fetchEndpoint<any>(
```
berndbohmeier: I would return the whole object here so that we can later use also different values returned by the relay.

cducrest:
This should be mergeable but it requires new version of the relay so I'll wait for it

cducrest:
Merged as part of other PR

PR #318, opened 2020-02-25
Add useful value for delegation fees in payments sending eth

cducrest:
Requires a PR in end2end repo to set gasPrice to 1000. I cannot push it because Github seems to bug. I'll try later.

### DIFF
--- a/src/Payment.ts
+++ b/src/Payment.ts
@@ -9,6 +9,7 @@ import { User } from './User'
 import utils from './utils'
 
 import {
+  DelegationFeesInternal,
   EventFilterOptions,
   FeePayer,
   isFeePayerValue,
@@ -120,7 +121,10 @@ export class Payment {
       )
       return {
         ethFees: utils.convertToAmount(ethFees),
-        delegationFees: utils.convertToDelegationFees(delegationFees),
+        delegationFees: utils.calculateDelegationFeesAmount(
+          delegationFees,
+          this.calculatePaymentGasLimit(path.length)
+        ),
         feePayer,
         maxFees,
         path,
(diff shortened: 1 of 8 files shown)


src/Payment.ts
```diff
@@ -298,4 +304,12 @@ export class Payment {
       path: result.path
     }
   }
+
+  private calculatePaymentGasLimit(pathLength: number) {
+    // Values taken from the contracts repository gas tests
+    const mediators = pathLength - 2
+    const transferBaseGas = 48000
+    const gasPerMediator = 18000
+    return (transferBaseGas + gasPerMediator * mediators) * GAS_LIMIT_MULTIPLIER
```
berndbohmeier: we should round that to an integer

src/Trustline.ts
```diff
@@ -180,9 +180,15 @@ export class Trustline {
         gasPrice: gasPrice ? new BigNumber(gasPrice) : undefined
       }
     )
+    // Value taken from contracts repository gas tests
+    // TODO: the gas limit for updating a TL could actually be lower than this depending on the situation
+    const MaxTrustlineUpdateGasLimit = 315_000 * GAS_LIMIT_MULTIPLIER
```
berndbohmeier: This should be rounded. Maybe put it into a function

src/utils.ts
```diff
@@ -271,6 +271,22 @@ export const convertToDelegationFees = (
...
 }
 
+export const calculateDelegationFeesAmount = (
+  delegationFees: DelegationFeesInternal,
+  gasLimit: number
+): Amount => {
+  const totalFeesRaw = delegationFees.baseFee.raw.plus(
+    delegationFees.gasPrice.raw
+      .multipliedBy(gasLimit)
+      .dividedToIntegerBy(1_000_000)
```
berndbohmeier: Name constant

src/Trustline.ts
```diff
@@ -256,9 +262,14 @@ export class Trustline {
         gasPrice: gasPrice ? new BigNumber(gasPrice) : undefined
       }
     )
+    // Value taken from contracts gas tests
+    const gasLimitCancelTrustlineUpdate = 19000 * GAS_LIMIT_MULTIPLIER
```
berndbohmeier: Round

PR #322, opened 2020-03-09
Feature/tx status

cducrest:
requires https://github.com/trustlines-protocol/relay/pull/455
Based on https://github.com/trustlines-protocol/clientlib/pull/312 for first commit

### DIFF
--- a/tests/e2e/Transaction.test.ts
+++ b/tests/e2e/Transaction.test.ts
@@ -1,0 +1,100 @@
+import chai from 'chai'
+import chaiAsPromised from 'chai-as-promised'
+import 'mocha'
+
+import BigNumber from 'bignumber.js'
+import { TLNetwork } from '../../src/TLNetwork'
+import { RawTxObject, TransactionStatus } from '../../src/typings'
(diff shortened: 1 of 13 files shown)


tests/e2e/Transaction.test.ts
```diff
@@ -0,0 +1,100 @@
...
+          )
+          await transaction.confirm(rawTx)
+          const txStatus = await transaction.getTxStatus(rawTx)
+
+          expect(txStatus.status).to.equal(TransactionStatus.Success)
+        })
+
+        if (testParameter.walletType === 'Identity') {
+          it('should get failed transaction status', async () => {
+            // Could not achieve to relay a failing transaction easily since the relay will reply with `Status 409 | There was an error while relaying this transaction`
```
berndbohmeier: But this should not be possible as well, we just did not implement the proper check yet, no? So we should maybe not have that test here or just have both tests testing for Failure or NotFound

cducrest: I am not sure what you mean. Do you think we should prevent relaying such meta-tx in the relay and then this tests will no longer work?

I guess I can test for NotFound, that may be better.

PR #325, opened 2020-03-13
Feature/delegation fees null when empty

cducrest:
Meta transactions now use the feeRecipient field filling it with what the relay gives
delegation fees will have undefined currencyNetwork when none are given by the relay.

goes with https://github.com/trustlines-protocol/relay/pull/457
closes https://github.com/trustlines-protocol/clientlib/issues/324

### DIFF
--- a/src/wallets/IdentityWallet.ts
+++ b/src/wallets/IdentityWallet.ts
@@ -341,6 +341,7 @@ export class IdentityWallet implements TLWallet {
       rawTx.gasPrice,
       rawTx.gasLimit
     )
+    rawTx.feeRecipient = rawTx.feeRecipient || metaTxFees.feeRecipient
     rawTx.currencyNetworkOfFees =
       rawTx.currencyNetworkOfFees || metaTxFees.currencyNetworkOfFees
 
--- a/src/wallets/IdentityWallet.ts
+++ b/src/wallets/IdentityWallet.ts
@@ -18,12 +18,15 @@ import {
   MetaTransactionFees,
   RawTxObject,
   Signature,
-  TransactionStatusObject
+  TransactionStatusObject,
+  TxFeesRaw,
+  TxObjectRaw
 } from '../typings'
 
 import utils from '../utils'
 
 import BigNumber from 'bignumber.js'
+import { AddressZero } from 'ethers/constants'
 
 // This is the proxy initcode without the address of the owner but with added 0s so that we only need to append the address to it
 const initcodeWithPadding =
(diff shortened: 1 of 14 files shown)


src/wallets/IdentityWallet.ts
```diff
@@ -341,6 +341,7 @@ export class IdentityWallet implements TLWallet {
       rawTx.gasPrice,
       rawTx.gasLimit
     )
+    rawTx.feeRecipient = rawTx.feeRecipient || metaTxFees.feeRecipient
```
berndbohmeier: Can it not be, that `metaTxFees.feeRecipient` is also undefined?
Should it not be `rawTx.feeRecipient || metaTxFees.feeRecipient || '0x000...'`

src/wallets/IdentityWallet.ts
```diff
@@ -341,6 +341,7 @@ export class IdentityWallet implements TLWallet {
...
     rawTx.currencyNetworkOfFees =
       rawTx.currencyNetworkOfFees || metaTxFees.currencyNetworkOfFees
```
berndbohmeier: Same here, I though metaTxFee.currenctyNetworkOfFees can not be null

src/wallets/IdentityWallet.ts
cducrest: I can add it, but it is already part of the relay (since the beginning I guess) that it has to return with feeRecipient: https://github.com/trustlines-protocol/relay/blob/master/src/relay/api/schemas.py#L76

src/wallets/IdentityWallet.ts
cducrest: `metaTxFee.currenctyNetworkOfFees` can be undefined. I thought that was what Daniel wanted to see.

berndbohmeier: but in the fees opject. If it is undefined in rawTx, this is an invalid transaction.

src/wallets/IdentityWallet.ts
berndbohmeier: Ah okay. Well to be complete I would say it should fill in the 0x0 address as default. At least this is what I would expect if the relay server would not specify a fee recipient.

src/wallets/IdentityWallet.ts
cducrest: Ok, I did it. It makes it impossible now to derive the fees from the rawTx as they have different interpretation of the `currencyNetworkOfFees` and `feeRecipient` values, so the signer is responsible for giving them both to `Transaction`.

If the relay provider replies with 0 fees with no currency network and fee recipient, then `rawTx` will have `zeroAddress` for these and `txFees` will have `undefined`.

PR #339, opened 2020-04-03
Feature/transfer info from event id

cducrest:
Based on https://github.com/trustlines-protocol/clientlib/pull/338
If I build the relay with https://github.com/trustlines-protocol/relay/pull/476, the e2e tests pass.


src/Payment.ts
```diff
@@ -317,17 +318,55 @@ export class Payment {
     )
   }
 
-  public async getTransferInformation(
+  public async getTransferInformationForTxHash(
```
berndbohmeier, line 321: Is the Information really adding something, why not just `getTransfer`?

src/Payment.ts
```diff
@@ -317,17 +318,55 @@ export class Payment {
...
+
+    return transferInformationList.map((transferInformation, index) =>
+      this.formatTransferInformationRaw(
+        transferInformation,
+        networkDecimalsList[index]
+      )
+    )
+  }
+
+  public async getTransferInformationForId(
```
berndbohmeier, line 352: Do we really have a reason to split them up?
Why not also have a function `getTransfer(event, options)`, where we use event id if present, and if not, fall back to transaction hash?
The reason would be that this is easier to maintain without breaking.

src/Payment.ts
berndbohmeier, line 321: Why do we actually need this function, to look up the Transfers by hash, now that we have the event_id?

src/Payment.ts
```diff
@@ -317,32 +318,108 @@ export class Payment {
...
-  public async getTransferInformation(
-    txHash: string,
+  /**
+   * Get the transfer details responsible for the transfer event with given log index and block number
+   * @param blockHash Block hash of the transfer event
+   * @param logIndex Log index of the transfer event
+   * @param options Optional network decimals for formatting the transfer value
+   */
+  public async getTransferDetails(
+    blockHash: string,
```
berndbohmeier, line 328: I think we misunderstood us, I thought here to have the same argument as in the other function. So one object that includes both. Otherwise this does not fulfill what I wanted, because the other function I dont care about.

berndbohmeier: Otherwise it should be good to go.

PR #349, opened 2020-04-14
Improve buildUrl function

### LINKED ISSUE #336: Add query parameters to generated links
Currently the clientlib generates URLs that have a predefined structure, but one is not able to add any additional query parameters to those links. 

Moreover the generated payment link is hard to work with as it can change it's path depending on whether one has provided a subject or not:
https://github.com/trustlines-protocol/clientlib/blob/master/src/Payment.ts#L263

We generate links in Payment, User and Contact. The createLink function in those classes is a proxy to utils.createLink, which calls utils.buildURL. 

I think that we should modify the buildUrl function and add a third param to it "queryParams". If the user needs additional params to the URL he could provide those as an array/object and the clientlib would create the correct link.

For example the Payment link could look like this
/:newtorkAddress/:userAddress/:amount?param1=1&param2=2 etc

The subject for a payment url should also be attached as a query param.

I don't think we can implement this as a non-breaking change as this would basically modify the signature of the functions… And in the case of createRequest I would actually change it from:

```
  public async createRequest(
    networkAddress: string,
    amount: number,
    subject: string,
    customBase?: string
  ): Promise<string> {
```
to

```
  public async createRequest(
    networkAddress: string,
    amount: number,
    queryParams: array|object,
    customBase?: string
  ): Promise<string> {
```

if this change is ok for you guys I could create a PR for this.

compojoom:
The buildURL function was expecting either an array or an object for pathParams. Depending on what it receives it generates a path URL or a query URL. That was not optimal in some situations.

I’ve changed the signature of the function to expect pathParams - array and queryParams - object. This way we can generate a url with both a path and query.

This is helpful in situations where users of the library would like to generate links to user account, contact or payment request but they also need to have some additional query params.

In the past the payment createRequest function was generating different paths depending on whether one provides a subject (it was difficult to match such paths and created a lot of overhead). Now the path is always the same:
/network/:network_address/:user_address/:amount if the user needs additional params to that url they are added as query params ?subject=1&note=note etc

fixes #336

### DIFF
--- a/src/Contact.ts
+++ b/src/Contact.ts
@@ -1,7 +1,7 @@
 import { TLProvider } from './providers/TLProvider'
 import { User } from './User'
 
-import utils from './utils'
+import utils, { defaultBaseUrl } from './utils'
 
 export class Contact {
   private user: User
@@ -20,15 +20,23 @@ export class Contact {
   /**
    * Creates sharable contact link.
    * @param address Address of contact to share.
-   * @param username Name of contact to share.
-   * @param customBase Optional custom base for link. Default `trustlines://`.
+   * @param options Optional
+   *        options[key] - additional params that will be appended
+   *        options.name - optional user name
+   *        options.subject - optional message for the recipient of the request
+   *        options.customBase - Optional custom base for link. Defaults to `trustlines://`.
    */
   public createLink(
     address: string,
-    username: string,
-    customBase?: string
+    options?: {
+      [key: string]: string
+      subject?: string
+      name?: string
+      customBase?: string
+    }
   ): string {
-    const params = ['contact', address, username]
-    return utils.createLink(params, customBase)
+    const path = ['contact', address]
+    const { customBase = defaultBaseUrl, ...rest } = options
+    return utils.buildUrl(customBase, { path, query: rest })
   }
 }
--- a/src/EthWrapper.ts
+++ b/src/EthWrapper.ts
@@ -176,7 +176,7 @@ export class EthWrapper {
     const { type, fromBlock } = filter
     const baseUrl = `tokens/${ethWrapperAddress}/users/${await this.user.getAddress()}/events`
     const events = await this.provider.fetchEndpoint<AnyTokenEventRaw[]>(
-      utils.buildUrl(baseUrl, { type, fromBlock })
+      utils.buildUrl(baseUrl, { query: { type, fromBlock } })
     )
     return events.map(event => utils.formatEvent(event, ETH_DECIMALS, 0))
   }
--- a/src/Payment.ts
+++ b/src/Payment.ts
@@ -10,7 +10,7 @@ import {
 } from './Transaction'
 import { User } from './User'
 
-import utils from './utils'
+import utils, { defaultBaseUrl } from './utils'
 
 import {
   DecimalsOptions,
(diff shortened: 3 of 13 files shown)


src/Contact.ts
```diff
@@ -22,13 +22,15 @@ export class Contact {
...
    * @param username Name of contact to share.
    * @param customBase Optional custom base for link. Default `trustlines://`.
+   * @param queryParams Optional query params for the generated url
    */
   public createLink(
     address: string,
     username: string,
-    customBase?: string
+    customBase?: string,
+    queryParams?: object
```
berndbohmeier, line 31: I would call it `params` or `options`, there is no query here.

berndbohmeier, line 31: Also two optional parameter are not good, it forces people to set a customBase when they only want to use `params`. 
We should handle it.

src/EthWrapper.ts
```diff
@@ -176,7 +176,7 @@ export class EthWrapper {
     const { type, fromBlock } = filter
     const baseUrl = `tokens/${ethWrapperAddress}/users/${await this.user.getAddress()}/events`
     const events = await this.provider.fetchEndpoint<AnyTokenEventRaw[]>(
-      utils.buildUrl(baseUrl, { type, fromBlock })
+      utils.buildUrl(baseUrl, [], { type, fromBlock })
```
berndbohmeier, line 179: We need to remove the need to set this `[]` if you dont need it

src/Payment.ts
```diff
@@ -261,23 +261,22 @@ export class Payment {
...
    * @param amount Requested transfer amount.
-   * @param subject Additional information for payment request.
+   * @param queryParams Additional query params for the generated url
    * @param customBase Optional custom base for link. Default `trustlines://`.
    */
   public async createRequest(
     networkAddress: string,
     amount: number,
-    subject: string,
+    queryParams?: object,
```
berndbohmeier, line 270: same here, should be `params` or `options` or so

berndbohmeier, line 270: Also order here is different from `createLink` in user

src/Payment.ts
```diff
@@ -261,23 +261,22 @@ export class Payment {
    * Creates a payment request link.
    * @param networkAddress Address of a currency network.
    * @param amount Requested transfer amount.
-   * @param subject Additional information for payment request.
+   * @param queryParams Additional query params for the generated url
```
berndbohmeier, line 264: We should give some documentation about known parameters, like subject, message, ..

src/utils.ts
```diff
@@ -127,41 +127,53 @@ export const websocketStream = (
...
  * Encodes URI components and returns a URL.
  * @param baseUrl base URL
- * @param params (optional) parameters for queries
+ * @param params (optional) path params for the url
+ * @param queryParams (optional) query params for the url
  */
-export const buildUrl = (baseUrl: string, params?: any[] | object): string => {
+export const buildUrl = (
+  baseUrl: string,
+  params?: any[],
```
berndbohmeier, line 135: we should not force people to use `params` to use `queryParams`. One way would be to check, what we got for each variable and just treat arras as `/sdf/sfd/` and an object as `?sdf=sf`

compojoom:
https://en.wikipedia.org/wiki/Uniform_Resource_Identifier
the call it component -> queryParams -> queryComponent ? Or maybe just query since the function name is buildURL, so query in this context should signify query component

src/Payment.ts
```diff
@@ -260,24 +260,28 @@ export class Payment {
   /**
    * Creates a payment request link.
    * @param networkAddress Address of a currency network.
-   * @param amount Requested transfer amount.
-   * @param subject Additional information for payment request.
-   * @param customBase Optional custom base for link. Default `trustlines://`.
+   * @param options
+   *        options.amount - optional amount for the payment request
```
berndbohmeier, line 264: Did we define if this is in base units or "normal", so if I request 1 euro, do I but there a 1 or 100000000

berndbohmeier: How do you use it atm?

Issue #344, opened 2020-04-14
Add payment request id to transfer


berndbohmeier:
Add an additional parameter to the prepare payment function to add a payment request id. As this could be used for different use cases, it might make sense to come up with a better name. 
Until we come up with an encoding, it is fine to just have a dummy encoding for now.

PR #352, opened 2020-04-15
Feature/add reference id for payments

weilbith:
Closes: #344 

The "reference id" is my proposal for a better naming as requested in the issue. It tries to be more generic by just stating that this is an identifier for something that the payment refers to.

The encoding of the id in addition to the client provided extra data is just a dummy function as asked for until there is an actual encoding format defined. So far it simply concatenates both strings.

### DIFF
--- a/src/Payment.ts
+++ b/src/Payment.ts
@@ -95,12 +95,17 @@ export class Payment {
     networkAddress: string,
     receiverAddress: string,
     value: number | string,
-    options: PaymentOptions = {}
+    options: PaymentOptions = {},
+    referenceId: string = ''
   ): Promise<PaymentTxObject> {
     const { gasPrice, gasLimit, networkDecimals, extraData } = options
     const decimals = await this.currencyNetwork.getDecimals(networkAddress, {
       networkDecimals
     })
+    const extraDataWithReferenceId = this.extendExtraDataByReferenceId(
+      extraData,
+      referenceId
+    )
     const { path, maxFees, feePayer } = await this.getTransferPathInfo(
       networkAddress,
       await this.user.getAddress(),
@@ -127,7 +132,7 @@ export class Payment {
           ),
           utils.convertToHexString(new BigNumber(maxFees.raw)),
           path,
-          extraData || '0x'
+          extraDataWithReferenceId
         ],
         {
           gasLimit: gasLimit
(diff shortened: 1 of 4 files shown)


src/Payment.ts
```diff
@@ -95,12 +95,17 @@ export class Payment {
     networkAddress: string,
     receiverAddress: string,
     value: number | string,
-    options: PaymentOptions = {}
+    options: PaymentOptions = {},
+    referenceId: string = ''
```
cducrest, line 99: I think it should be part of the options

src/Payment.ts
```diff
@@ -95,12 +95,17 @@ export class Payment {
...
   ): Promise<PaymentTxObject> {
     const { gasPrice, gasLimit, networkDecimals, extraData } = options
     const decimals = await this.currencyNetwork.getDecimals(networkAddress, {
       networkDecimals
     })
+    const extraDataWithReferenceId = this.extendExtraDataByReferenceId(
```
cducrest, line 105: A more generic name would be more convenient when we will need to add more in the future, like `encodeExtraData`

src/Payment.ts
cducrest: I also think it could better be named `receiverReference` since `referenceId` could also mean the reference we will use to bound a message to a payment.

weilbith: The reason I did not included into the options was that the `PaymentOptions` are used at multiple places e.g. prepare to close a trustline. Do you think it still fits into that interface or would it require a new interface extending the `PaymentOptions`? :thinking:

src/Payment.ts
weilbith: Hmm I'm not sure what this means. Is this then a generic function which takes an extraData string plus a dictionary of all possible values to set? So far it is a very focused function to just encodes the given reference. So since the encoding is totally unclear at the moment, I would not know how to rework this function without making assumptions for the signature. While the current one do no assumptions at all.
If you have a concrete idea in mind and I just stuck here, please don't hesitate to make a suggestion or add a commit by yourself. I would be glad if so.

src/Payment.ts
berndbohmeier: as we can not come up with a better name, I suggest we stick to `paymentRequestId`

berndbohmeier: It has to go in the options. If it does not fit the type, change the type/create a new one.

weilbith: :heavy_check_mark:

berndbohmeier:
LGTM, but lets not merge it yet, as I first want to release the new clientlib version.

PR #359, opened 2020-04-22
Use MessagePack to encode extraData

### LINKED ISSUE #358: Use msgpack to encode extraData

berndbohmeier:
Use MessagePack to encode the extraData of a Transfer. This is so far
only the paymentRequestId. I chose to simply use bytes as the datatype
for now, but that might change later, and we might want to use a custom
type.
The attaching of the decoded paymentRequestId had to be moved, so that
it will be encluded in every endpoint where Transfers are received.
At the moment it can not handle the old extraData and paymentRequestId
at the same time. This is due to the unclear behaviour of what extraData
on the receiver side should contain, if both were included.

Closes #358

### DIFF
--- a/src/Event.ts
+++ b/src/Event.ts
@@ -2,6 +2,7 @@ import { Observable } from 'rxjs/Observable'
 import { fromPromise } from 'rxjs/observable/fromPromise'
 
 import { CurrencyNetwork } from './CurrencyNetwork'
+import { decode } from './extraData'
 import { TLProvider } from './providers/TLProvider'
 import { User } from './User'
 
@@ -81,7 +82,9 @@ export class Event {
       )
     ])
     return events.map(event =>
-      utils.formatEvent<T>(event, networkDecimals, interestRateDecimals)
+      processExtraData(
+        utils.formatEvent<T>(event, networkDecimals, interestRateDecimals)
+      )
     )
   }
 
@@ -100,7 +103,9 @@ export class Event {
     const events = await this.provider.fetchEndpoint<AnyEventRaw[]>(
       parameterUrl
     )
-    return this.setDecimalsAndFormat(events)
+    return (await this.setDecimalsAndFormat(events)).map(event =>
+      processExtraData(event)
+    )
   }
 
   /**
@@ -125,7 +130,13 @@ export class Event {
             return this.currencyNetwork
               .getDecimals(event.networkAddress)
               .then(({ networkDecimals, interestRateDecimals }) =>
-                utils.formatEvent(event, networkDecimals, interestRateDecimals)
+                processExtraData(
+                  utils.formatEvent(
+                    event,
+                    networkDecimals,
+                    interestRateDecimals
+                  )
+                )
               )
           } else {
             return Promise.resolve(event)
@@ -239,4 +250,20 @@ export class Event {
       return result
     }, {})
   }
+}
+
+/**
+ * Processes the content of extraData and attaches the content to the event.
+ * @param event
+ */
+function processExtraData(event) {
+  if (event.extraData) {
+    const extraData = decode(event.extraData)
+    if (extraData) {
+      if (extraData.paymentRequestId) {
+        event.paymentRequestId = extraData.paymentRequestId
+      }
+    }
+  }
+  return event
 }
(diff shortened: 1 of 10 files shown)


src/extraData.ts
```diff
@@ -0,0 +1,54 @@
+import { decode, encode } from '@msgpack/msgpack'
+
+export const MSG_PACK_ID = '0x544c4d50' // hex encoded "TLMP" (Trustlines Message Pack)
+
+export interface ExtraData {
```
cducrest, line 5: Why do we put this here and not in `typing.ts`

src/extraData.ts
```diff
@@ -0,0 +1,54 @@
+import { decode, encode } from '@msgpack/msgpack'
+
+export const MSG_PACK_ID = '0x544c4d50' // hex encoded "TLMP" (Trustlines Message Pack)
```
weilbith, line 3: Chance to save a comment. :shrug:

```suggestion
export const MSG_PACK_ID = binascii.hexlify(b"TLMP")
```

cducrest:
I would prefer if we used a custom type directly, because I found it hard to understand why we need to convert the `paymentRequestId` to a byte array. https://github.com/trustlines-protocol/clientlib/pull/359/files#diff-598410b6dd753780b7924a1e7b5fbc1fR20

src/extraData.ts
berndbohmeier: If you have that function in javascript for me, I am all in.

weilbith: Oh dear :joy:

src/Event.ts
```diff
@@ -81,7 +82,9 @@ export class Event {
       )
     ])
     return events.map(event =>
-      utils.formatEvent<T>(event, networkDecimals, interestRateDecimals)
+      handleExtraData(
+        utils.formatEvent<T>(event, networkDecimals, interestRateDecimals)
+      )
```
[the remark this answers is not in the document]
berndbohmeier: I dont know what you mean by that. Thats how the language works, there is afaik no pipeline operator like
`event -> handleExtraData | utils.formatEvent`
I might put it in a function though.

src/extraData.ts
berndbohmeier: Because it belongs here and not in `typing.ts`. Having a central file is wrong in many ways, most importantly it supports tight coupling of normally separated components. 
Types should be as local as the involving code, to check that these types are  compatible with other parts of the code is the job of the typechecker.

berndbohmeier: I think that extraData handling here was so isolated from the rest, that I think it is fine to break the (wrong) convention of using `typings.ts`

berndbohmeier:
> I would prefer if we used a custom type directly, because I found it hard to understand why we need to convert the paymentRequestId to a byte array. https://github.com/trustlines-protocol/clientlib/pull/359/files#diff-598410b6dd753780b7924a1e7b5fbc1fR20

I dont think this is related. We have to convert it because I want to use the bytes type to save gas. This would stay the same if I want to use a custom type X, we then have to convert it to X before. 
We could change it in a way that it automatically recognizes hex strings as bytes, or X, but that is possible for both approaches and would have the problem that someone could not start easily a string with `0x` 
Custom types changes the way people have to work with the encoded data, I dont think that local code structure are a good argument there to make that decision. 
If you have an argument why a custom type is better, let me know.

berndbohmeier:
I am open to change this, but as I think it is easy to do this in a later PR, I will just merge it as is.

src/Event.ts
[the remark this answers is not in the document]
berndbohmeier: I think it is fine if we handle this later, if it is not readable enough.

cducrest:
>  We have to convert it because I want to use the bytes type to save gas. This would stay the same if I want to use a custom type X, we then have to convert it to X before.


That's what I understood, the reason we use `hexToArray` here is totally different from the reason we use `hexToArray` or `arrayToHex` elswhere in the code. I would say this should be extracted, and would have been extracted if we used a custom type, but you are correct that is wrong thinking.

berndbohmeier:
I am not aware that we use that elsewhere in the code.

src/Event.ts
```diff
@@ -240,3 +251,15 @@ export class Event {
     }, {})
   }
 }
+
+function handleExtraData(event) {
```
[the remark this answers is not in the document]
berndbohmeier: I changed it to `processExtraData` and put a doc string. Hope that helps a bit at least.

Issue #369, opened 2020-05-04
Decode extra data from payment information


cducrest:
When getting transfer information we get extra data that was in transfer, we should probably decode it.
https://github.com/trustlines-protocol/clientlib/blob/master/src/Payment.ts#L503

PR #371, opened 2020-05-05
Decode extra data of transfer details

cducrest:
Always return paymentRequestId and messageId for transfer details
and transfer events, return null when there is none instead of
undefined.

closes https://github.com/trustlines-protocol/clientlib/issues/369

### DIFF
--- a/src/Payment.ts
+++ b/src/Payment.ts
@@ -2,7 +2,11 @@ import { BigNumber } from 'bignumber.js'
 
 import { CurrencyNetwork } from './CurrencyNetwork'
 import { Event } from './Event'
-import { encode as encodeExtraData } from './extraData'
+import {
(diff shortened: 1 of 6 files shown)


src/Payment.ts
```diff
@@ -527,3 +528,16 @@ function encode({
     return '0x'
   }
 }
+
+function processExtraData(transferDetails): TransferDetails {
```
weilbith, line 532: This is basically 99% the same as we already have in `Events.ts`. They are called the same, basically do the same and both were affected by this PR in the same manner (hint that the functionality is actually tight). Could we make this a single function and maybe put it as utility to `extraData.ts`?

cducrest: makes sense

Issue #361, opened 2020-04-27
use random nonce for meta-tx


cducrest:
Use random nonce > 2**255 instead of asking nonce from the relay.
Do not remove current solution, but leave it for later as an option.

PR #373, opened 2020-05-07
Generate meta transaction nonces randomly

weilbith:
Closes: #361

I'm not too happy that the `nonce` for the `RawTxObject` interface is now accepting `number | string`. As I talked to Come this could/should be a `string` always. But it feels like this a too big change for this PR, focusing on something different.

### DIFF
--- a/src/wallets/IdentityWallet.ts
+++ b/src/wallets/IdentityWallet.ts
@@ -32,6 +32,10 @@ import { AddressZero } from 'ethers/constants'
 const initcodeWithPadding =
   '0x608060405234801561001057600080fd5b506040516020806102178339810180604052602081101561003057600080fd5b50506101d6806100416000396000f3fe6080604052600436106100295760003560e01c80635c60da1b1461005c578063d784d4261461008d575b600080546040516001600160a01b0390911691369082376000803683855af43d6000833e808015610058573d83f35b3d83fd5b34801561006857600080fd5b506100716100c2565b604080516001600160a01b039092168252519081900360200190f35b34801561009957600080fd5b506100c0600480360360208110156100b057600080fd5b50356001600160a01b03166100d1565b005b6000546001600160a01b031681565b6000546001600160a01b03161561014957604080517f08c379a000000000000000000000000000000000000000000000000000000000815260206004820152601a60248201527f496d706c656d656e746174696f6e20616c726561647920736574000000000000604482015290519081900360640190fd5b600080546001600160a01b03831673ffffffffffffffffffffffffffffffffffffffff19909116811790915560408051918252517f11135eea714a7c1c3b9aebf3d31bbd295f7e7262960215e086849c191d45bddc9181900360200190a15056fea165627a7a7230582009e814f2e28666ad200a2e809c1fe802a9264c378c39d6c2032706ab340c09a40029000000000000000000000000'
 
+const MIN_RANDOM_NONCE = new BigNumber(2).pow(255).plus(1)
+const MAX_RANDOM_NONCE = new BigNumber(2).pow(256)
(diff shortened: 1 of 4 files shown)


src/wallets/IdentityWallet.ts
```diff
@@ -480,3 +484,22 @@ export function calculateIdentityAddress(
...
+  const random = BigNumber.random()
+  const nonce = random.multipliedBy(RANDOM_NONCE_RANGE).plus(MIN_RANDOM_NONCE)
+  const coefficient = nonce.c
+  const exponent = nonce.e
+  const nonceAsString = coefficient.reduce(
+    (nonceString, coefficientPart) =>
+      nonceString.concat(coefficientPart.toString()),
+    ''
+  )
+  return nonceAsString.substr(0, exponent + 1)
```
cducrest, line 504: Why can you not just do `nonce.toString()`?

src/wallets/IdentityWallet.ts
```diff
@@ -480,3 +484,22 @@ export function calculateIdentityAddress(
...
+
+/**
+ * Generates a random nonce to use for meta transactions.
+ * The nonce fits into the range of ]2^255, 2^256[.
+ * This is an alternative to the up counting nonce (]0, 2^255[) without the need
+ * to query a [[TLProvider]].
+ */
+export function getRandomNonce(): string {
+  const random = BigNumber.random()
+  const nonce = random.multipliedBy(RANDOM_NONCE_RANGE).plus(MIN_RANDOM_NONCE)
```
cducrest, line 496: The range is size `2**255`, you should use enough randomness so that the full capacity of the range is used: `random = BigNumber.random(77)`. I chose 77 because `2*255 < 10**17`

see: https://mikemcl.github.io/bignumber.js/#random

cducrest, line 496: You should make sure the nonce is an integer: `nonce.integerValue(BigNumber.ROUND_DOWN)`

cducrest, line 496: You should add 1 to this, I am not sure what you do with the reduce later, maybe you do it there.
The random number generated can be `0`, in which case the nonce is `MIN_RANDOM_NONCE` which is too low.
However if you add `1`, the nonce generated can be `2**255 - 1 + 2*255 + 1 = 2**256` which is too high, so you should lower the range by `1`.

I would suggest using `MIN_RANDOM_NONCE = new BigNumber(2).pow(255).plus(1)`

src/wallets/IdentityWallet.ts
weilbith: Due to [the docs](https://mikemcl.github.io/bignumber.js/#exponential-at) it allows to configure the `EXPONENTIAL_AT` value for the `toString()` function only up to `1e+9`, which is not sufficient for us.

src/wallets/IdentityWallet.ts
weilbith: Should not matter how it is processed afterwards. Correct?

weilbith: Yes, this is correct. I obviously missed that even I knew.

src/wallets/IdentityWallet.ts
weilbith: As we discussed in person I used the configuration options wrongly during my
experiments so that is whole thing can be just simplified. Will apply that.

src/wallets/IdentityWallet.ts
weilbith: I'll do since I'll remove the awkward string generation (see [this
discussion](https://github.com/trustlines-protocol/clientlib/pull/373#discussion_r421555230)).

PR #382, opened 2020-07-01
Add method for creating a trustline request invite link

dakingha69:
This method creates a link, which can be shared off-chain, inviting the receiver to create a trustline request with the specified parameters.

The app wants to use these links, for example, to create dynamic links and improve the onboarding experience for new users.

Related to https://github.com/trustlines-network/ppm-app-project/issues/317

### DIFF
--- a/src/Trustline.ts
+++ b/src/Trustline.ts
@@ -10,7 +10,7 @@ import {
 } from './Transaction'
 import { User } from './User'
 
-import utils from './utils'
+import utils, { defaultBaseUrl } from './utils'
 
 import {
   AnyNetworkTrustlineEvent,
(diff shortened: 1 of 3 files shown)


src/Trustline.ts
```diff
@@ -600,6 +600,53 @@ export class Trustline {
...
+    const path = [
+      'trustlinerequest',
+      networkAddress,
+      String(creditlineGiven),
+      String(creditlineReceived),
+      String(interestRateGiven),
+      String(interestRateReceived)
+    ]
+
+    return utils.buildUrl(customBase || defaultBaseUrl, { path, query: rest })
```
cducrest, line 647: Feels a bit weird that we import `defaultBasUrl` from utils to give it back to a utils function to use as default.
I guess this should be the default behaviour of the function if `customBase` is undefined.

dakingha69: Yes agree. Changed `utils.buildUrl`

PR #385, opened 2020-07-08
Update readme

cducrest:
Update the readme to follow the documentation layout described in https://github.com/trustlines-network/project/issues/877

I don't know if we should add a `pre-requisite` section to the `Get Up and Running` section, if so what should be in there?
I am not sure how to write the `Release` section as I was never sure I was doing it properly. Can you do that @dakingha69 ?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,78 @@
+# Contributing
+
+Welcome and thank you for wanting to contribute to the Trustlines Project.
+We appreciate all help, depending on your knowledge, this can be
+spreading the word, helping with documentation, reporting issues or even contributing to the source code.
+
+Feel free to checkout the code, play around with it and run the tests.
+Before making a change, please first discuss the change via opening an issue on this
+repository. If you just need some small question answered, you can also reach out to
+us via our [gitter chat](https://gitter.im/trustlines/community).
+
+## Open an issue
+
+To request a feature or report a bug, please open an issue on the issue tracker in github.
+
+### Report a bug
+
+Please include a short description of the problem, including what you expected to happen and what did happen instead.
+Please also include the version of the clientlib you were using, your operating system and the used configuration.
+
+### Request a feature
+
+Please include a description of what you would like to see implemented and an explanation of why you believe this would
+be a good addition.
+
+## Open a pull request
+
+### Before starting
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -3,14 +3,24 @@
...
+and a mobile application can be built on top of it.
 
-### Install via package manager
+The goal of the clientlib is to provide a user focused means to interract with Trustlines via a relay server.
+The intent is to make it as easy as possible to build a user interface for Trustlines on top of it.
 
-We publish the trustlines-clientlib to npm as a ES6 module.
+## Get Up and Running
+
+[// todo write pre-requisites, is it obvious already when we say es6 module?]: #
```
dakingha69, line 20: Yes, I think it we can omit that. It should be quite obvious and other JS libs I looked at also do not really specify the pre-requisites.

CONTRIBUTING.md
```diff
@@ -0,0 +1,78 @@
...
+  - Fix the bug, so that the test passes.
+- For a new feature:
+  - Add a test, that demonstrate the new feature.
+  - Add or update related documentation if necessary.
+- Add an entry to the unreleased section of the [changelog](CHANGELOG.md).
+- Format your code changes with prettier. If you setup pre-commit, this will happen automatically on commit.
+
+## Changelog
+
+We loosely follow [keep a changelog](https://keepachangelog.com/en/0.3.0/).
```
berndbohmeier, line 48: On this repo we follow this actually more strict, so you might want to change the wording here from `add` to `added` for example, and they are only grouped, so do not have to start with the keyword.

README.md
```diff
@@ -3,14 +3,24 @@
...
+The clientlib is a component of the [Trustlines Protocol](https://trustlines.foundation/protocol.html).
+The Trustlines Protocol is a set of rules to allow the transfer of value on top of existing trust
+relationships stored on a trustless infrastructure, here a blockchain.
 
-## Installation
+In the technology stack, the clientlib is located on top of a [relay server](https://github.com/trustlines-protocol/relay),
+and a mobile application can be built on top of it.
 
-### Install via package manager
+The goal of the clientlib is to provide a user focused means to interract with Trustlines via a relay server.
```
berndbohmeier, line 15: I would add, that it also provides the wallet including key handling.

PR #390, opened 2020-08-20
Add function in identity wallet to update identity implementation

cducrest:
Is it enough for the mobile app to have a function in the identity wallet to do this, or does it need a function in something like `User` or `TLNetwork`?

How do we want the user to know whether he should update or not (whether he already has the same implementation address on the contract on chain and in the identityWallet of the clientlib)?
We can either make a relay endpoint to ask for the value of `identityImplementation` for the identity proxy, or we can save the last used value of identity implementation somehow in the app.

closes https://github.com/trustlines-network/project/issues/884

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "reconnecting-websocket": "^4.1.10",
     "rxjs": "^5.4.1",
     "simple-jsonrpc-js": "0.0.10",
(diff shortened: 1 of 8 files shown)


cducrest:
Merged in https://github.com/trustlines-protocol/clientlib/pull/391

Issue #192, opened 2019-02-12
Optimize webpack bundle


dakingha69:
Besides creating a `commonjs` and `esmodule`, we also create a UMD bundle using webpack. The bundle size is 1.61 MB though and should be reduced.

cducrest:
Maybe reduce the size of `contracts.json` by filtering the contracts that end up in it.

cducrest:
Maybe check if ethers is too big and newer versions are smaller?

PR #393, opened 2020-09-21
Add script that filters the packed abi to make bundle smaller

cducrest:
It reduces the size of `_bundles` from 1332kB to 1004kB.
I am not really sure if that is clean, or if that is all that is needed for https://github.com/trustlines-protocol/clientlib/issues/192?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: circleci/node:10.15
+    - image: circleci/node:14.11
 
 commands:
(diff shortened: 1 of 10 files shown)


dakingha69 (approved):
In general looks good. I don't know if the bundle size could also be further reduced somehow. For example, you can use tools like https://github.com/webpack-contrib/webpack-bundle-analyzer which outputs something like:
image

Maybe it is possible to properly use tree shaking with webpack and other things which I am currently not aware of. But a bundle size of over 1 MB is still to big I am afraid for closing https://github.com/trustlines-protocol/clientlib/issues/192

Issue #190, opened 2019-02-11
Change user.address to user.getAddress


dakingha69:
To compatible with web3 signer strategy and additional potential strategies the attribute `address`  in `User` should be replaced by `signer.getAddress()` as defined in the `TLSigner` interface. Note that this change is a breaking change.

cducrest:
The `getAddress()` is defined, we need to remove the legacy `address` function.

PR #394, opened 2020-09-21
Update address used in User.ts to getAddress()

cducrest:
This makes the methods `createLink` and `requestEth` work with web3providers.

I decided not to remove the `address` method of User, it will raise an error if called when using a web3provider just as many other
method do (e.g. User.showSeed), and provides for a convenient way to access the user address when you know you are dealing with a local wallet.

I don't mind deciding the opposite if you think that's better.

closes: https://github.com/trustlines-protocol/clientlib/issues/190

### DIFF
--- a/src/User.ts
+++ b/src/User.ts
@@ -39,7 +39,7 @@ export class User {
   }
 
   /**
-   * Checksummed Ethereum address of currently loaded user/wallet.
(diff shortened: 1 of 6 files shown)


dakingha69 (approved):
LGTM 👍 Yea, I do not have strong objections on leaving the address getter.

PR #396, opened 2020-09-23
Add documentation to getAll function of event to filter by contract

cducrest:
Adds documentation to query parameter to filter events in `getAll`, see: https://github.com/trustlines-protocol/relay/pull/542

Using this filter in the app can potentially speed up the querying of event, depending on whether the event database has events from other type of contracts.

### DIFF
--- a/src/typings.ts
+++ b/src/typings.ts
@@ -107,6 +107,7 @@ export interface Amount {
 // EVENTS
 export interface EventFilterOptions {
   type?: string
+  contractType?: 'CurrencyNetwork' | 'Exchange' | 'UnwETH' | 'Token'
   fromBlock?: number
 }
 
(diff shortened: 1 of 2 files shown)


src/typings.ts
```diff
@@ -107,6 +107,7 @@ export interface Amount {
 // EVENTS
 export interface EventFilterOptions {
   type?: string
+  contractType?: string
```
dakingha69, line 110: You could enforce the usage of valid contract types here. Either using string literals
```ts
type ValidContractTypes =  'CurrencyNetwork' | 'Exchange' | 'UnwETH' | 'Token'
```
or using an enum.

dakingha69 (approved):
Looks good. Small suggestion on typing the valid contract types.

Issue #266, opened 2019-11-18
Function to query collected fees in between two timestamps


cducrest:
Also write e2e tests, see https://github.com/trustlines-protocol/contracts/blob/master/tests/test_information_from_events.py for test data

PR #399, opened 2020-10-19
Add function to get earned mediation of user in network

cducrest:
Need https://github.com/trustlines-protocol/relay/pull/547 to provide the endpoint to work
Closes https://github.com/trustlines-protocol/clientlib/issues/266

### DIFF
--- a/src/UserInformation.ts
+++ b/src/UserInformation.ts
@@ -7,6 +7,10 @@ import {
   AccruedInterestsObject,
   AccruedInterestsRaw,
   DecimalsOptions,
+  EarnedMediationFeesListObject,
+  EarnedMediationFeesListRaw,
+  MediationFeeObject,
+  MediationFeeRaw,
   TrustlineAccruedInterestsObject,
   TrustlineAccruedInterestsRaw,
   UserAccruedInterestsObject,
@@ -16,9 +20,9 @@ import {
 import { CurrencyNetwork } from './CurrencyNetwork'
 
 /**
- * The Interests class contains methods related to getting accrued interests for trustlines or users.
+ * The UserInformation class contains methods related to getting trustlines related information of a user.
  */
-export class Interests {
+export class UserInformation {
   private provider: TLProvider
   private user: User
   private currencyNetwork: CurrencyNetwork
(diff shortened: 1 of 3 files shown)


src/typings.ts
```diff
@@ -686,6 +686,37 @@ export interface TrustlineAccruedInterestsObject {
...
+  network: string
+  mediationFees: MediationFeeRaw[]
+}
+
+export interface EarnedMediationFeesListObject {
+  user: string
+  network: string
+  mediationFees: MediationFeeObject[]
+}
+
```
compojoom, line 719: So, why do we need both Raw and Object definitions if they are 100% the same?

compojoom: Ah, `value: Amount` is the difference. hm...

src/UserInformation.ts
```diff
@@ -33,6 +37,39 @@ export class Interests {
...
   }
 
+  public async getUserEarnedMediationFees(
+    networkAddress: string,
+    options: {
+      timeWindowOption?: { startTime: number; endTime: number }
+      decimalsOptions?: DecimalsOptions
+    } = {}
+  ): Promise<EarnedMediationFeesListObject> {
+    const baseUrl = `networks/${networkAddress}/users/${await this.user.getAddress()}/mediation-fees`
```
compojoom, line 47: I would put this await this.user.getAddress() on a separate line. I know it is annoying to type that and to come up with yet another variable, but readability comes first. 
Right now just by flying over this code one can miss that we have an async call on this line since the await is hidden inside of the template literal.

compojoom (changes requested):
Looks OK.
It is a breaking change since you are renaming the class. If the class is now only supposed to give userinformation and no longer trustline information, then there is no point of having methods that are called `getUserEarnedMediationFees`, `getUserAccruedInterests`. We could just call those `getEarnedMediationFees`and `getAccruedInterests`

the `getAccruedInterest" is the total interest the user has accrued accross all trustlines with no timeframe? If that is the case why not call it getTotalAccruedInterest.

`getTrustlineAccruedInterests` since we are refactoring, what do you think about calling this `getInterestsAccruedOnTrustline`?

src/typings.ts
cducrest: One of them is what you get from the relay server, the other one is what you provide to the user, with interpretation of the decimals to use to show the amount.

cducrest:
I renamed the methods as you suggested. I did not rename `getAccruedInterests` to `getTotalAccruedInterest` because it can be used with an optional timeframe. Also `total` make it sound like a sum while it is a list.

compojoom:
awesome 👍 

What is with the failing e2e tests on circle?

cducrest:
As written in the first comment, it `Need trustlines-protocol/relay#547 to provide the endpoint to work`.
This is just using an endpoint that is implemented in pr 547 of the relay.

compojoom (approved):

PR #403, opened 2020-11-20
Add remainingData to the encoding of payment ExtraData

cducrest:
To prepare a payment, you can now provide `remainingData` as arbitrary data that will be encoded into the `extraData` that will end up on the chain. 
You can still ignore this field, together with `paymentRequestId` and `TransferId` and provide custom `extraData`.
This `remainingData` will also be returned in events and such.
Call me if that's too ambiguous.

closes https://github.com/trustlines-network/project/issues/859


dakingha69 (approved):
LGTM 👍

PR #406, opened 2020-12-04
Add support for payment request cancellation

nekhaly:
Adding method paymentRequestCancel that allows sending a message to
counterparty to cancel a pre-sent payment request.
Message will have PaymentRequestCancel type and includes payment
request id to be cancelled.

### DIFF
--- a/src/Messaging.ts
+++ b/src/Messaging.ts
@@ -107,17 +107,20 @@ export class Messaging {
   /**
    * Sends a payment request cancel message to given `counterParty` and returns created message.
    * @param counterPartyAddress Address of counter party.
-   * @param id id of the payment request to decline
+   * @param id id of the payment request to cancel
    * matches either the nonce as a number or id of a payment request as a hex string.
+   * @param subject Optional subject of cancel message.
    */
   public async paymentRequestCancel(
     counterPartyAddress: string,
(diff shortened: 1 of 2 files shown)


cducrest:
This is useless, you can use `paymentRequestDecline` and check on the receiver end if the id corresponds to a sent payment or a received payment to know if someone is cancelling a payment request you sent or they sent.
But if you want it, it's fine.

Add the `subject` parameter for consistency though.

nekhaly:
> This is useless, you can use `paymentRequestDecline` and check on the receiver end if the id corresponds to a sent payment or a received payment to know if someone is canceling a payment request you sent or they sent.
> But if that makes your life easier we can have it.
> 
> Add the `subject` parameter for consistency though.

Just to archive the conclusion of what we were talking about. The assumption I built the solution about is that `decline` and `cancel` are different functionalities and that they may diverge in the future and would need to be loosely coupled. The tradeoff to have repeated code that contradicts with DRY principle has been made for the sake of another principle of SOC. We may discuss again in a common discussion about `client-lib` to determine its role more accurately, and this I think would help in determining the pathway for it to be more generalized or specialized.

cducrest (approved):

PR #424, opened 2021-09-27
Fix trustline updates with transfer of "0"

compojoom:
If we are triyng to update a trustline and we pass a transfer value of "0"(string) the clientlib will call the overloaded updateTrustline function of the contract. This causes the meta transaction to fail as there is already an existing trustline and when updating we shouldn't be passing a transfer value.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,7 +95,8 @@ jobs:
       DOCKER_REPO: trustlines/e2e
       LOCAL_IMAGE: e2e
     steps:
-      - setup_remote_docker
+      - setup_remote_docker:
+          version: 20.10.7
       - checkout
       - run:
           name: Build docker image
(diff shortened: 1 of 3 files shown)


cducrest:
I think this is wrong, you don't allow for strings that are over the number limit for javascript. I would say we need to use something like calcRaw and then compare with bigNumber(0) https://github.com/trustlines-protocol/clientlib/blob/8b30ce108a7e5ec6e90ff3359c2c248ee8a330d4/src/utils.ts#L170

compojoom:
yes, yes. You are right. That was what I thought on my way to work this morning... I'll have to figure something else.

cducrest (approved):

cducrest (approved):

cducrest (approved):

